PR #2, opened 2018-01-16
Drop intltool

jtojnar:
> gettext has seen quite some enhancements in recent years, after Daiki Ueno started maintaining it. It can now extract (and merge back) strings from diverse file formats, including many of the formats that are important for desktop applications. With gettext 0.19.8, there is really no need anymore to use intltool or GLib’s dated gettext glue (AM_GLIB_GNU_GETTEXT and glib-gettextize). 

https://wiki.gnome.org/Initiatives/GnomeGoals/GettextMigration


sthibaul:
Pulled, thanks!

PR #5, opened 2018-01-25
modules: Add initial support for the Kali French/English speech synthesizer

sthibaul:


sthibaul:
It seems github really doesn't want to let me view the 500-line diff... Thus writing here:

There is a problem when calling GetNomVoixKali(): the buffer size is not passed, so it could overflow.
We're looking about this issue with Kali upstream.

dusek:
> It seems github really doesn't want to let me view the 500-line diff...

Speaking of `src/modules/kali.cpp` in the “Files changed” tab, I got the 495-line diff displayed fine by clicking the “Load diff” link.

sthibaul:
> I got the 495-line diff displayed fine by clicking the “Load diff” link.

That's what I clicked in my browser indeed, but nothing showed up (I actually forgot about it, it saying "Loading" for like an hour)

PR #37, opened 2018-07-20
Remove pyxdg dependency

hadess:
And use the same algorithm as GLib to determine the base user config
directory (XDG_CONFIG_HOME, then ~/.config then /tmp/$USERNAME/.config)

### DIFF
--- a/src/api/python/speechd_config/config.py.in
+++ b/src/api/python/speechd_config/config.py.in
@@ -30,8 +30,6 @@ import socket
 import sys
 import time
 
-from xdg import BaseDirectory
-
 # Locale/gettext configuration
 
 locale.setlocale(locale.LC_ALL, '')
@@ -167,7 +165,16 @@ class Tests:
 
     def user_conf_dir(self):
         """Return user configuration directory"""
-        return os.path.join(BaseDirectory.xdg_config_home, "speech-dispatcher")
+        config_dir = os.environ['XDG_CONFIG_HOME']
+        if not config_dir:
+            home_dir = os.environ['HOME']
+            if home_dir:
+                config_dir = os.path.join(home_dir, ".config")
+            else:
+                tmpdir = os.environ['TMPDIR'] or "/tmp/"
+                config_dir = os.path.join(tmpdir, os.getlogin(), ".config")
+
+        return os.path.join(config_dir, "speech-dispatcher")
 
     def system_conf_dir(self):
         """Determine system configuration directory"""


hadess:
> Mmm, is it really a problem to have the pyxdg dependency?

In our case, yes, speechd is only one of 4 packages that required pyxdg. This patch, and another one to another package, means there are only 2 packages that now require pyxdg.

This particular patch is already merged into Fedora, and will likely be carried there and in derivatives.

> It's better to use libraries which are supposed to implement standards correctly than re-implement it.

I agree, but this is something that's specified, and well-defined, so +7 lines of code to not have to rely on another package is a tiny cost. If the daemon already relied on GLib through introspection, I would have used that instead.

> I guess the actual issue being fixed here is the fallback to /tmp/$USERNAME/.config? I'd say it should rather be put into python-xdg than in speech-dispatcher. That way it will also benefit a lot other use cases.

I didn't do any "bug fixing" here, I just implemented the functionality the same way GLib does it, which might be more resilient than the version in pyxdg.

sthibaul:
hadess, le mer. 25 juil. 2018 04:56:48 -0700, a ecrit:
>     Mmm, is it really a problem to have the pyxdg dependency?
> 
> In our case, yes, speechd is only one of 4 packages that required pyxdg.

Which case is it?

> This patch, and another one to another package, means there are only 2
> packages that now require pyxdg.
> 
> This particular patch is already merged into Fedora, and will likely be carried
> there and in derivatives.

But again, why trying to kill pyxdg?

>     It's better to use libraries which are supposed to implement standards
>     correctly than re-implement it.
> 
> I agree, but this is something that's specified, and well-defined,

Perhaps specified and well-defined (see below), but $DEITY knows how the
standard will evolve in the future, and it's much better to maintain
an implementation in a well-known library than copy/paste the code in
various places.

> so +7 lines of code to not have to rely on another package is a tiny
> cost.

The package itself is not many KiB. Is it really worth the (my!)
maintenance effort for that?

>     I guess the actual issue being fixed here is the fallback to /tmp/$USERNAME
>     /.config? I'd say it should rather be put into python-xdg than in
>     speech-dispatcher. That way it will also benefit a lot other use cases.
> 
> I didn't do any "bug fixing" here, I just implemented the functionality the
> same way GLib does it, which might be more resilient than the version in pyxdg.

Which precisely means that the standard is not so much "specified and
well-defined". Notably the /tmp fallback is not specified by the
standard.

Which again shows that it's better to have a shared implementation than
copy/pasting one, only to realize that there is yet another thing to
fix, and having to propagate the fix everywhere.

hadess:
> Which case is it?

RHEL

> But again, why trying to kill pyxdg?

Because I don't fancy maintaining a library for 10 years when it can be replaced by 7 lines of code :)

sthibaul:
hadess, le mer. 25 juil. 2018 05:10:42 -0700, a ecrit:
>     But again, why trying to kill pyxdg?
> 
> Because I don't fancy maintaining a library for 10 years when it can be
> replaced by 7 lines of code :)

So you are instead asking dozens of projects (in Debian 132 projects
depend on it) to maintain these 7 lines of code instead of sharing the
maintenance?

If it was 7 lines of code whose behavior is never to change, I'd agree,
but as the difference of implementation between glib and pyxdg shows,
the behavior may change even without a standard change. For instance
TMPDIR is not defined within windows and /tmp does not make sense there,
so the glib fallback is not complete either.

hadess:
> So you are instead asking dozens of projects (in Debian 62 packages
> depend on it) to maintain these 7 lines of code instead of sharing the
> maintenance?

No, I asked just this one because I don't want to maintain pyxdg for 10 years just to get a directory path. I'm sure there are a number of applications that use pyxdg in ways that can't be replaced by a couple of lines of code, that's not the case here.

> For instance
> TMPDIR is not defined within windows and /tmp does not make sense there,
> so the glib fallback is not complete either.

pyxdg doesn't support Windows. GLib's implementation (`g_get_user_config_dir`) does:
https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-get-user-config-dir

> Yes, putting config files for speech-dispatcher in a directory whose content could disappear after a reboot is completely unacceptable.

When you get to that point, it means you don't have a $HOME, and this would be the least of your worries.

sthibaul:
hadess, le mer. 25 juil. 2018 05:33:27 -0700, a ecrit:
>     For instance
>     TMPDIR is not defined within windows and /tmp does not make sense there,
>     so the glib fallback is not complete either.
> 
> pyxdg doesn't support Windows.

But it could be made to.

> GLib's implementation (g_get_user_config_dir) does:
> [1]https://developer.gnome.org/glib/stable/
> glib-Miscellaneous-Utility-Functions.html#g-get-user-config-dir

Yes, but the snippet you propose does not support that.

It's precisely that kind of small fixes I'm thinking of: having them
shared within python-xdg means less maintenance work overall. Yes, it
means one package more for distributions to maintain, but one source of
issues less to maintain for applications.

hadess:
> But it could be made to.

None of the code in speechd seems to support Windows, and the server uses Unix-only functions, so I don't think it's fair to take potential future Windows support into account when judging this patch.

> Yes, but the snippet you propose does not support that.

As I said, nothing in the codebase seems to support Windows right now, but I'll happily rework the patch to use GLib directly. Do you want to add a runtime dependency on GLib's introspection instead then?

sthibaul:
hadess, le mer. 25 juil. 2018 07:04:37 -0700, a ecrit:
>     But it could be made to.
> 
> None of the code in speechd seems to support Windows,

Sure, it's not my point.

My point is that removing python-xdg completely means copy/pasting a few
lines of code into several dozens of software while it does need fixing
as the example shows. speech-dispatcher is not the only one among them.

> Do you want to add a runtime dependency on GLib's introspection
> instead then?

No, python-xdg is just what speech-dispatcher needs.

PR #44, opened 2018-08-19
Convert old #ifdefs for Solaris into autoconf checks

alanc:
Years ago, Willie Walker contributed fixes for building on Solaris 10 using #ifdef checks for the Solaris compilers and GNU libc.

This pull request converts these to autoconf checks so that you can build on Solaris with either gcc or Solaris compilers, and so that when building on current Solaris releases you don't redefine functions that Solaris 10 did not have but Solaris 11 now does.  (It should also improve portability to other non-Linux systems.)

The -lsocket -lnsl checks could be simplified by using AC_SEARCH_LIBS, as is done in https://gitlab.freedesktop.org/xorg/lib/libxtrans/blob/master/xtrans.m4#L28-30 - but then the results would just be in $LIBS, not split out into $EXTRA_SOCKET_LIBS as was done before, so
I didn't do that.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -73,8 +73,8 @@ AC_TYPE_UINT32_T
 AC_FUNC_FORK
 AC_FUNC_MALLOC
 AC_FUNC_REALLOC
-AC_CHECK_FUNCS([daemon dup2 gethostbyname gettimeofday memmove memset mkdir])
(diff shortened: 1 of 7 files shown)


sthibaul:
Thanks!

PR #72, opened 2018-11-02
Fix NEWS etc. for the release

sthibaul:


alexarnaud:
FYI, there is a little typo here on the NEWS file:
Better manage volukme for generic-based modules

Best regards,
Alex.

PR #84, opened 2018-11-06
baratinoo: Add support for SSML say-as tag

cwendling:
Add translation from SSML to Baratinoo's proprietary format for the `<say-as>` tag.  It's pretty similar, and the same set of attributes are supported.

### DIFF
--- a/src/modules/baratinoo.c
+++ b/src/modules/baratinoo.c
@@ -1149,6 +1149,24 @@ static void ssml2baratinoo_start_element(GMarkupParseContext *ctx,
 		int i = attribute_index(attribute_names, "level");
 		g_string_append_printf(state->buffer, "\\emph<{%s}",
 				       i < 0 ? "" : attribute_values[i]);
+	} else if (strcmp(element, "say-as") == 0) {
+		int i_as = attribute_index(attribute_names, "interpret-as");
+		int i_fmt = attribute_index(attribute_names, "format");
+		int i_detail = attribute_index(attribute_names, "detail");
+
+		if (i_as < 0) {
+			DBG(DBG_MODNAME "Missing required 'interpret-as' attribute of '<say-as>' tag");
+			i_fmt = i_detail = -1;
+		} else if (i_fmt < 0 && i_detail >= 0) {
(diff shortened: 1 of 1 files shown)


sthibaul:
I guess the extra spaces when not all three values are specified don't pose problem?

sthibaul:
Mmm, actually they are perhaps even needed to avoid ambiguity?

cwendling:
No, they are neither needed nor pause a problem.  The documentation simply states they are optional, and my test with the actual engine shows it simply doesn't care about the extra 2 spaces if those attributes are not set.

sthibaul:
Reading the Baratinoo tags ref manual, it seems they are optional, but probably in order only.
I would say that it'd be good to introduce DBG warnings in case

- i_as is not specified (it's really mandatory), in which case the tag should be emitted, but without parameters at all (just setting i_fmt and i_detail to -1).
- i_detail is specified but i_fmt is not specified, in which case i_detail should be ignored (set it to -1), to avoid letting baratinoo take if_detail for a i_fmt

so that in case we end up with these, we get a chance to see that in the debugging logs.

sthibaul:
(yes, emitting the tag without parameters is odd, but at least it'll keep the matching with \sayas<

cwendling:
@sthibaul I could do that.  However, I just pushed a version no supporting the `detail` attributes, because it's indeed problematic for the order and the engine's SSML documentation specifies it ignores the `detail` attribute anyway.

> (yes, emitting the tag without parameters is odd, but at least it'll keep the matching with `\sayas<`

Yeah exactly the reason why I didn't actually check whether the `intepret-as` value was set or not before emitting the tag.

sthibaul:
Ok, let's drop detail for now then (but keep a comment about that in the code).

> Yeah exactly the reason why I didn't actually check whether the `intepret-as` value was set or not before emitting the tag.

Please still set `i_fmt` to -1 when `i_as <0`, so we don't risk baratinoo takes a `fmt` as an `as`.

sthibaul:
(and emit a DBG warning)

cwendling:
@sthibaul hehe, our messages are crossing ^^
I just implemented your initial suggestion keeping all attributes but performing the sanity checks and emitting the debugging messages.

Issue #169, opened 2018-11-21
baratinoo: assertion failure/crash related to new marks


cwendling:
I'm not sure yet what leads to this, but there is a serious bug in the current Baratinoo module, which can in some cases lead to a Pulse (in my case) assertion failure apparently due to a race condition/bad mutex:
```
Assertion 'pa_atomic_load(&(s)->_ref) >= 1' failed at pulse/stream.c:336, function pa_stream_get_state(). Aborting.
```
or sometimes
```
Assertion 'pthread_mutex_destroy(&m->mutex) == 0' failed at pulsecore/mutex-posix.c:83, function pa_mutex_free(). Aborting.
Assertion 'pa_atomic_load(&(c)->_ref) >= 1' failed at pulse/context.c:1063, function pa_context_get_state(). Aborting.
```
or even a nice glibc backtrace:
```
*** Error in `/usr/lib/speech-dispatcher-modules/sd_baratinoo': corrupted double-linked list: 0x00007f1148000f20 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x70bfb)[0x7f116bf2abfb]
/lib/x86_64-linux-gnu/libc.so.6(+0x76fc6)[0x7f116bf30fc6]
/lib/x86_64-linux-gnu/libc.so.6(+0x7733d)[0x7f116bf3133d]
/lib/x86_64-linux-gnu/libc.so.6(+0x78dfa)[0x7f116bf32dfa]
/lib/x86_64-linux-gnu/libc.so.6(__libc_calloc+0xb6)[0x7f116bf359b6]
/lib64/ld-linux-x86-64.so.2(+0xb2d6)[0x7f116ed292d6]
/lib64/ld-linux-x86-64.so.2(+0x587d)[0x7f116ed2387d]
/lib64/ld-linux-x86-64.so.2(+0x880c)[0x7f116ed2680c]
/lib64/ld-linux-x86-64.so.2(+0x13bd4)[0x7f116ed31bd4]
/lib64/ld-linux-x86-64.so.2(+0xf704)[0x7f116ed2d704]
/lib64/ld-linux-x86-64.so.2(+0x136c9)[0x7f116ed316c9]
/lib/x86_64-linux-gnu/libc.so.6(+0x11f0dd)[0x7f116bfd90dd]
/lib64/ld-linux-x86-64.so.2(+0xf704)[0x7f116ed2d704]
[...]

PR #170, opened 2018-11-22
module_tts_output_marks() fixes

cwendling:
Fix detecting the marks order, and properly report all marks.

Part of #169.


cwendling:
I just force-pushed a fix, because I apparently didn't build after this version using `DBG()` without braces around, and actually `DBG()` is a tricky macro.  Fixed now.

---
PS: it could be nice to wrap `DBG()`'s body in `do { } while (0)` to avoid that kind of trickiness and allow what I had.

PR #215, opened 2019-03-27
build: drop spdincludedir

jengelh:
speech-dispatcher's include strategy is such:

include/fdsetconv.h:#include <speechd_types.h>
clients/say/say.c:#include <libspeechd.h>

(sbl does the same.)

This mandates that the include file is placed at exactly
${includedir} and not any subdirectory.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -464,12 +464,6 @@ AC_SUBST([spdlibdir])
 audiodir="$spdlibdir"
 AC_SUBST([audiodir])
 
-# Path for speech-dispatcher include files:
-spdincludedir=${includedir}/speech-dispatcher
-AC_SUBST([spdincludedir])
-includedir=${spdincludedir}
-AC_SUBST([includedir])
-
 # support for systemd unit files
 AC_ARG_WITH([systemdsystemunitdir],
         AS_HELP_STRING([--with-systemdsystemunitdir=DIR], [Directory for systemd service files]),
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -19,6 +19,6 @@
 
 noinst_HEADERS = fdsetconv.h spd_utils.h i18n.h safe_io.h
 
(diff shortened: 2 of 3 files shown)


sthibaul:
Well, the speech-dispatcher.pc does set includedir to ${prefix}/include/speech-dispatcher, so the abovementioned #includes will just work.

Moving the speech-dispatcher files out of the speech-dispatcher/ directory would be contrary to the move initiated by 7a217e8ef5fcae2cdf3541f326e961982742fc43 to avoid polluting /usr/include/ with all speech-dispatcher include files.

Which actual problem are you getting? Perhaps you are just not using speech-dispatcher.pc as you should be?

jengelh:
>to avoid polluting /usr/include/ with all speech-dispatcher include files.

The `#include` namespace is still "polluted"/"occupied" the same way it was, so moving files around in the _filesystem_ does not really buy anything. To really address that, the directory should be part of the include name, similar to how the X11 include files do it:

```
// someprogram.c
#include <speech-dispatcher/libspeechd.h>
```

```
// /usr/include/speech-dispatcher/libspeechd.h
#include <speech-dispatcher/libspeechd_version.h>
#include <speech-dispatcher/speechd_types.h>
```

(Such a change would also give the opportunity to compact the names, e.g. to <speech-dispatcher/types.h>)

sthibaul:
> The #include namespace is still "polluted"/"occupied" the same way it was

Right.

> the directory should be part of the include name

Ok, that approach looks fine to me.

> give the opportunity to compact the names

That'd break all existing programs, including those which actually already use `#include <speech-dispatcher/libspeechd.h>`, so I'd rather avoid it.

sthibaul:
@jengelh : any news on this?

jengelh:
No news. I am not working on it, and there are no plans to pick it up again.

PR #260, opened 2019-06-10
New symbols importation; emojis and unicode font variants support

### LINKED ISSUE #49: Emojis support
Emojis are more and more used and for the moment, they are hard to read for blind users on Linux.
Implementing emojis support on speechDispatcher could allow to have them supported in every voice supported synthesisers.
A begin of solution could be to use http://cldr.unicode.org/ for the emoji list as NVAccess did for NVDA issue 6523 (nvaccess/nvda#6523) as the emojis list.

### LINKED ISSUE #217: Add support for unicode font variants
Step to reproduce: In a terminal, execute: "spd-say 𝓯𝓸𝓸"

Expected result: Something intelligible would be presented.

Actual result: "Letter 1d4ef, letter 1d4f8, letter 1d4f8"

Impact: These font variants cannot be presented in a sane fashion by any client of speech-dispatcher (including, but not limited to, Orca).

Rather than having every speech-dispatcher client do the translation, it would be great if speech-dispatcher could handle this on behalf of all clients. Thanks in advance!

pzajda:
Add a script to import symbols dictionnary and a make import-symbols into locale directory.
The script is in locale/import-symbols.py
The  following dictionaries are imported:
* NVDA symboles dictionaries from NVDA sources;
* Unicode CLDR characters to add emojis support (locale/<lang>/emojis.dic), fixes #49
* Unicode font variants from UnicodeData.txt (locale/en/font-variants.dic), fixes #217

Last NVDA symbols dictionaries and font-variants.dic have been imported.


locale/Makefile.am
```diff
@@ -20,55 +20,107 @@
...
+	cs/emojis.dic \
 	da/symbols.dic \
+	da/emojis.dic \
 	de/symbols.dic \
+	de/emojis.dic \
 	el/symbols.dic \
+	el/emojis.dic \
 	en/symbols.dic \
+	en/emojis.dic \
+	en/font-variants.dic \
```
sthibaul: I'd say put this one at the top of the list, to make it more apparent

locale/an/symbols.dic
```diff
@@ -1,22 +1,17 @@
-#Aragonese locale/ar/symbols.dic
+﻿# This file is generated automatically by the script import-symbols.py during make
```
sthibaul: Thinking about it... I'd say do not put the details in the file itself, but in a README.md file in locale/ . In the files, just mentioned that they are generated automatically, and mention the README.md file.

locale/Makefile.am
```diff
@@ -20,55 +20,107 @@
 nobase_localedata_DATA = \
```
sthibaul: I believe that CLDR has more than these? I'd say put all emojis files, even if we don't have a corresponding NVDA file.

locale/import-symbols.py
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/env python3.5
+# coding: utf-8
+# NVDA symbols and cldr import
+# Import NVDA symbols dictionaries then emojies from CLDR database
+# Copyright (C) 2001, 2019 Brailcom, o.p.s.
```
sthibaul: I  believe you write this file completely? Then put your own copyright, not Brailcom's

locale/import-symbols.py
```diff
@@ -0,0 +1,186 @@
...
+	# We brows the source files
+	for source in sources:
+		tree = ElementTree.parse(source) # We parse the CLDR XML file to extract all emojies
+		for element in tree.iter("annotation"):
+			if element.attrib.get("type") == "tts":
+				cldrDict[element.attrib['cp']] = element.text.replace(":","")
+	assert cldrDict, "cldrDict is empty"
+	return cldrDict
+
+# The list of languages supported by Speech-Dispatcher and the corresponding language(s) in the CLDR database
```
sthibaul: There is not really any "list of languages supported by Speech-Dispatcher", it's up to the speech syntheses. Whenever espeak-ng adds a new language, it'll immediately work in speech-dispatcher, without any change there. So we shouldn't have any list, and just pick put everything we can from CLDR, e.g. af, am, etc. (af for instance does happen to be supported by espeak-ng)

sthibaul: And the same for NVDA symbols: there is no reason to avoid importing the NVDA symbols file even if CLDR doesn't provide emoji names. Just import everything that is availble from NVDA.

pzajda:
Sorry for the strange log, an amend which has happened very strangely.
To resume, a major rewrite:
* Process NVDA symbols and CLDR dictionaries separatly
* In dictionaries headers, specify to read README.md in locale
* Group exceptions for CLDR files
* Also import languages which not exist yet
* Created locale/README.md

locale/import-symbols.py
```diff
@@ -65,81 +65,38 @@ def createCLDRAnnotationsDict(sources):
...
-	"ar":("ar",),
-	"bg":("bg",),
-	"ca":("ca",),
-	"cs":("cs",),
-	"da":("da",),
-	"de":("de",),
-	"el":("el",),
+# List of languages which are non-standard for CLDR files
+CLDRExceptionsList = {
 	"en":("en_001","en"),
```
sthibaul: I think we want to take `en.xml` from CLDR? `en_001.xml` seems way smaller

locale/import-symbols.py
```diff
@@ -65,81 +65,38 @@ def createCLDRAnnotationsDict(sources):
...
-	"sr":("sr",),
-	"sv":("sv",),
-	"ta":("ta",),
-	"tr":("tr",),
-	"uk":("uk",),
-	"vi":("vi",),
-	"zh":("zh",),
+	"sr":("sr","sr_Cyrl","sr_Cyrl_BA","sr_Latn","sr_Latn_BA"),
+	"yue": ("yue","yue_Hans"),
 	"zh_CN":("zh",),
```
sthibaul: I believe we want to keep zh_CN as zh_CN, and not make it zh. There isn't really a "default" zh localization.

locale/import-symbols.py
```diff
@@ -65,81 +65,38 @@ def createCLDRAnnotationsDict(sources):
...
-	"is":("is",),
-	"it":("it",),
-	"ja":("ja",),
-	"ka":("ka",),
-	"kn":("kn",),
-	"ko":("ko",),
-	"lt":("lt",),
-	"my":("my",),
+	"es_US":("es_419","es_US"),
 	"nb_NO":("nb",),
```
sthibaul: I don't think there is any reason to keep this `nb_NO` exception. Having `nb_NO` and no `nb` will be fine

locale/import-symbols.py
```diff
@@ -65,81 +65,38 @@ def createCLDRAnnotationsDict(sources):
...
-	"nl":("nl",),
 	"nn_NO":("nn",),
```
sthibaul: Same as `nb_NO`

locale/import-symbols.py
```diff
@@ -65,81 +65,38 @@ def createCLDRAnnotationsDict(sources):
...
-	"pa":("pa",),
-	"pl":("pl",),
 	"pt_BR":("pt",),
```
sthibaul: I don't think pt_BR will be a good "pt" file. Really keep it as "pt_BR" for Brazil, and "pt_PT" will be fine for Portugal.

locale/import-symbols.py
```diff
@@ -65,81 +65,38 @@ def createCLDRAnnotationsDict(sources):
...
-	"hu":("hu",),
-	"is":("is",),
-	"it":("it",),
-	"ja":("ja",),
-	"ka":("ka",),
-	"kn":("kn",),
-	"ko":("ko",),
-	"lt":("lt",),
-	"my":("my",),
+	"es_US":("es_419","es_US"),
```
sthibaul: Like for `en`, `es_419.xml` looks small compared to `es.xml`, so I don't see a reason to use the former instead of the latter.

sthibaul:
I have merged it, many thanks!

PR #270, opened 2019-11-14
Improve locale make and add documentation for newly added dictionaries

pzajda:
The "symbolsrc" directory should be ignored by git as it is added manually when importing symbols and emojis dictionaries. I propose to add the directory to the GITIGNOREFILES variable in locale/Makefile.am.

I also noticed I forgot to add new emojis dictionaries to this make file in the PR importing last updated dictionaries.
To prevent avoiding this happening in the future, I suggest to add some precision about what to do when new dictionaries are imported when following instructions in locale/README.md


sthibaul:
Thanks!

PR #285, opened 2020-01-19
Update the sd_ibmtts module for compatibility with voxin 3.0

gcasse:
sd_ibmtts can still be built without voxin installed using the '--with-ibmtts=shim' configure option.
voxin.h is included and copied from libvoxin 1.4.6 (https://github.com/Oralux/libvoxin).
    
* Note for voxin users
The voxin installer removes sd_ibmtts and uses sd_voxin instead.
sd_voxin simply helps to select according to the version of speech-dispatcher a suitable sd_ibmtts module customized for voxin 3.0 ( https://github.com/Oralux/voxin-installer/tree/master/src/speechd-voxin )
    
* Example 
Tests using voxin 3.0 + IBM TTS (American_English) and Vocalizer Embedded (nathan-embedded-compact)

Example 1: ask the ibmtts module to list voices
Command: LC_ALL=C spd-say -o ibmtts -L
    
                         NAME                 LANGUAGE                  VARIANT
             American_English                    en-US                     none
      nathan-embedded-compact                       en                       US
    
Example 2: ask the ibmtts module to say a sentence
Command: LC_ALL=C spd-say -y nathan-embedded-compact -o ibmtts "Hello Mr Smith"

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -230,8 +230,7 @@ AC_ARG_WITH([ibmtts],
 	[with_ibmtts=check])
 AS_IF([test $with_ibmtts != "no"],
 	[AC_CHECK_LIB([ibmeci], [eciStop],
-		[with_ibmtts=yes;
-		ibmtts_include="-I/opt/IBM/ibmtts/inc/"],
+		[with_ibmtts=yes],
 		[AS_IF([test $with_ibmtts = "yes"],
 			[AC_MSG_FAILURE([IBMTTS is not available])])
 		 with_ibmtts=shim])])
@@ -240,7 +239,6 @@ if test "$with_ibmtts" = shim -a $enable_shared = no; then
 fi
 AM_CONDITIONAL([ibmtts_support], [test $with_ibmtts != no])
 AM_CONDITIONAL([ibmtts_shim], [test $with_ibmtts = shim])
-AC_SUBST([ibmtts_include])
 AS_IF([test $with_ibmtts != no], [output_modules="${output_modules} ibmtts"])
 
 # check for ivona support
--- a/src/modules/ibmtts.c
+++ b/src/modules/ibmtts.c
@@ -50,9 +50,14 @@
 #include <string.h>
 #include <glib.h>
 #include <semaphore.h>
+#include <dlfcn.h>
+#include <ctype.h>
 
 /* IBM Eloquence Command Interface. */
 #include <eci.h>
+
+/* Libvoxin include */
+#include <voxin.h>
 
 /* Speech Dispatcher includes. */
 #include "spd_audio.h"
@@ -240,11 +245,12 @@ static SPDVoice **ibmtts_voice_list = NULL;
 static int *ibmtts_voice_index = NULL;
 
 /* Internal function prototypes for main thread. */
+static void ibmtts_update_sample_rate();
 static void ibmtts_set_language(char *lang);
 static void ibmtts_set_voice(SPDVoiceType voice);
 static char *ibmtts_voice_enum_to_str(SPDVoiceType voice);
 static void ibmtts_set_language_and_voice(char *lang, SPDVoiceType voice,
-					  char *dialect);
+					  char *dialect, char *name);
 static void ibmtts_set_synthesis_voice(char *);
 static void ibmtts_set_rate(signed int rate);
 static void ibmtts_set_pitch(signed int pitch);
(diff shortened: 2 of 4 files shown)


src/modules/ibmtts.c
```diff
@@ -1332,11 +1452,12 @@ static void ibmtts_set_synthesis_voice(char *synthesis_voice)
 
 	DBG("Ibmtts: %s, synthesis voice=%s", __FUNCTION__, synthesis_voice);
 
-	for (i = 0; i < MAX_NB_OF_LANGUAGES; i++) {
-		if (!strcasecmp(eciLocales[i].name, synthesis_voice)) {
+	for (i = 0; (i < VOX_MAX_NB_OF_LANGUAGES) && eciLocales[i].name; i++) {
+	  if (!strcasecmp(eciLocales[i].name, synthesis_voice)) {
```
sthibaul: In the whole source code, indentation uses tabs, not spaces.

src/modules/ibmtts.c
```diff
@@ -1698,43 +1819,86 @@ static char *ibmtts_search_for_sound_icon(const char *icon_name)
...
 	int i = 0;
 
-	if (eciGetAvailableLanguages(aLanguage, &nLanguages))
+	/* if voxGetVoices available, update the list of installed voices */
+	if (_voxGetVoices) {
+		number_of_voices = VOX_RESERVED_VOICES;
+		if (!_voxGetVoices(voices, &number_of_voices) && (number_of_voices <= VOX_RESERVED_VOICES)) {
+			int i, j;
+			int min_id = eciLocales[VOX_ECI_VOICES-1].langID;			
+			for (i=0, j=VOX_ECI_VOICES; i<number_of_voices; i++) {
```
sthibaul: It'd be better to avoid hardcoding that we will only have VOX_RESERVED_VOICES languages provided by voxin. That limit will be broken sooner or later.

Better:

* Keep `eciLocales` as it is in the current git master, just rename it to `eciHardcodedLocales`
* Introduce a global `eciLocale *eciLocales = eciHardcodedLocales;` pointer, which thus points by default to `eciHardcodedLocales`.
* When `voxGetvoices` is available, call it once with `NULL` to get the number of voices, then dynamically allocate a new `eciLocale` array of the proper size, and fill it.

That way, ibmtts will either use the hardcoded array (old voxin), or the dynamically-provided array (new voxin) without distorted array indexes etc.

All that being said, I don't understand why using `voxGetVoices`: AIUI voxin is just a forwarder for ibmtts, that is not actually necessary: the speech-dispatcher plugin can just directly link against ibmtts, and use ibmtts-provided functions. Here, won't the `eciGetAvailableLanguages` call simply just work already? If not yet, can't that be solved instead of introducing yet more complexity? AFAICT, `voxGetVoices` ends up just calling `eciGetAvailableLanguages` anyway, so I don't see the benefit of calling the former rather the latter, but see the drawback: it prevents from linking again ibmtts directly.

src/modules/ibmtts.c
```diff
@@ -1698,43 +1819,86 @@ static char *ibmtts_search_for_sound_icon(const char *icon_name)
...
+				eciLocale *local = eciLocales + j;
+				size_t len = 0;
+				vox_t *vox = voices + i;
+				if (vox->id <= min_id) /* id already known? */
```
sthibaul: With the dynamically-allocated array, I wouldn't bother trying to eliminate the already-known langages. Possibly voxin would revamp the ids someday and that'd screw everything up. Since long-run-wise it provides a function that just gives the set of language, I'd say just use that and forget about the hardcoded information.

src/modules/voxin.h
```diff
@@ -0,0 +1,60 @@
...
+#define LIBVOXIN_VERSION_MINOR 4
+#define LIBVOXIN_VERSION_PATCH 6
+
+typedef enum {voxFemale, voxMale} voxGender;
+typedef enum {voxAdult, voxChild, voxSenior} voxAge;
+
+#define VOX_ECI_VOICES 22
+#define VOX_RESERVED_VOICES 30
+#define VOX_MAX_NB_OF_LANGUAGES (VOX_ECI_VOICES + VOX_RESERVED_VOICES)
+#define VOX_STR_MAX 128
```
sthibaul: Such kind of hardcoded sizes always pose problem: they either

* are too big and thus eat memory for no use
* are too small and thus truncates names etc.

Better use pointers to dynamically-allocated strings than inlining arrays in the structure.

configure.ac
```diff
@@ -230,8 +230,7 @@ AC_ARG_WITH([ibmtts],
 	[with_ibmtts=check])
 AS_IF([test $with_ibmtts != "no"],
 	[AC_CHECK_LIB([ibmeci], [eciStop],
-		[with_ibmtts=yes;
-		ibmtts_include="-I/opt/IBM/ibmtts/inc/"],
```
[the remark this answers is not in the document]
gcasse: Thanks for your review. 
As time allows, this monolithic patch will be resent, split in smaller pieces, and will include most of your notices regarding this speech-dispatcher module.

gcasse:
This pull request is replaced by #377 "Add voxin module", which takes in account the upstream review above.

PR #297, opened 2020-04-13
Fix symbols importation script and update dictionaries

Nardol:
* import-symbols.py: call to env python3 instead of env python3.5
* Update CLDR to version 36.1
* Update NVDA symbols dictionaries to last beta branch version


sthibaul:
Thanks!

PR #342, opened 2020-05-16
Return "NOT YET IMPLEMENTED" from all history commands.

jpwhiting:
Most history commands already give not yet implemented response
from what I see in testing locally. however history get client_list
still gives a list of all clients which is not good security wise.
For now always return not yet implemented to all history commands.
At some point we can add some security measures to only give
history for the current client, etc. but just remove all history
until then.

### DIFF
--- a/src/server/parse.c
+++ b/src/server/parse.c
@@ -273,6 +273,9 @@ enddata:
 char *parse_history(const char *buf, const int bytes, const int fd,
 		    const TSpeechDSock * speechd_socket)
 {
+	return g_strdup(ERR_NOT_IMPLEMENTED);
+
+#if 0
 	char *cmd_main;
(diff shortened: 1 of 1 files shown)


sthibaul:
Well, I could have done that myself :) (and I'll probably do that before the 0.10 release).

It shouldn't be too hard to just drop the part of parse_history that provides a client list and allows to look into their history.

jpwhiting:
Yep, good point, ok will do.

sthibaul:
Thanks :) Also, please update doc/ssip.texi, currently in @subsection Message History in SSIP it says that for security reasons only history of the same uid is allowed, this needs to be fixed into saying that even the same uid is not enough, for flatpak etc. reasons

jpwhiting:
Second stab here: https://github.com/brailcom/speechd/pull/343

PR #356, opened 2020-07-09
pulse: handle _pulse_open() errors in reconnect

wtay:
When the sample format changes and we try to reopen our pulseaudio
connection, check for errors or else we end up passing a NULL to
pw_simple_write() and cause an abort.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1770778


sthibaul:
Thanks!

PR #377, opened 2020-08-30
Add voxin module

gcasse:
This module is compatible with Voxin >= 3.0.

### DIFF
--- a/src/modules/voxin.c
+++ b/src/modules/voxin.c
@@ -1,0 +1,1955 @@
+/*
+ * voxin.c - Speech Dispatcher backend for Voxin (version >= 3)
+ *
+ * Copyright (C) 2006, 2007 Brailcom, o.p.s.
+ * Copyright (C) 2020 Gilles Casse <gcasse@oralux.org>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Based on ibmtts.c:
+ * @author  Gary Cramblitt <garycramblitt@comcast.net> (original author)
+ *
+ * $Id: ibmtts.c,v 1.30 2008-06-30 14:34:02 gcasse Exp $
+ */
+
+/* This output module operates with four threads:
+   
+   The main thread called from Speech Dispatcher (module_*()).
+   A synthesis thread that accepts messages, parses them, and forwards
+   them to Voxin.
+   This thread receives audio and index mark callbacks from
+   IBM TTS and queues them into a playback queue. See _ibmtts_synth().
+   A playback thread that acts on entries in the playback queue,
+   either sending them to the audio output module (module_tts_output()),
+   or emitting Speech Dispatcher events.  See _ibmtts_play().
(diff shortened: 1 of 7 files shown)


sthibaul:
Hello,

Errr, I don't think we want to replicate almost all of the ibmtts.c source code like this. We already have such a thing between espeak and espeak-ng, and that has already posed questions, that's not the way we want to see the spd code go. See for instance https://github.com/brailcom/speechd/pull/373 which you do not have in your version, while I guess you want it. Sharing the implementation really is what we want to make it maintainable without headaches.

I have now integrated various changes you have made, notably the more flexible language selection. While looking at the difference between your driver and the current ibmtts, I am still left with the question which is still unanswered:

“I don't understand why using voxGetVoices: AIUI voxin is just a forwarder for ibmtts, that is not actually necessary: the speech-dispatcher plugin can just directly link against ibmtts, and use ibmtts-provided functions. Here, won't the eciGetAvailableLanguages call simply just work already? If not yet, can't that be solved instead of introducing yet more complexity? AFAICT, voxGetVoices ends up just calling eciGetAvailableLanguages anyway, so I don't see the benefit of calling the former rather the latter, but see the drawback: it prevents from linking again ibmtts directly.”

Is the current ibmtts.c not already working with voxin 3.0? What poses problem?

src/modules/voxin.c
```diff
@@ -0,0 +1,1955 @@
...
+	if (old_index == new_index) {
+		DBG(DBG_MODNAME "LEAVE %s, no change", __FUNCTION__);
+		return;
+	}
+
+	language = voices[new_index].lang;
+	region = voices[new_index].variant;
+
+	/* Fix locale name for French Canadian */
+	if (!strcmp(language, "ca") && !strcmp(region, "FR")) {
```
sthibaul, line 1887: Also, ewwww...
AIUI it's just a bug in voxin that you fixed in 451ba54dff0c84f731aede9e045ed2a0b0831d9f

sthibaul:
Just to be clear: we can build the same source file twice with different `-D` parameters so we can use ifdefs and keep just one source file to maintain, that's probably what we'll do with espeak vs espeak-ng.

gcasse:
Hello,

Before 2016, sd_ibmtts directly called e.g. the
eciGetAvailableLanguages() function from the IBM TTS 32 bits library.

Since January 2016, libvoxin implements the ECI API to originally ease
the integration of the 32 bits IBM TTS binaries in a 64 bits environment
for various accessibility clients (sd_ibmtts, emacspeak outloud server,...).

Today, sd_voxin is built for x86_64, armh, aarch64 (ARM only with the
Vocalizer Embedded voices).

gcasse:
The strcmp line has been removed from my speechd fork but unfortunately
remained in this pull request.
Good catch!

sthibaul:
> Today, sd_voxin is built for x86_64, armh, aarch64 (ARM only with the Vocalizer Embedded voices).

Does this mean that there are arm versions of the ECI engine binary? Or is the original binary run through simulation?

gcasse:
sd_voxin on ARM is used for the Vocalizer Embedded voices.

sthibaul:
> sd_voxin on ARM is used for the Vocalizer Embedded voices.

I don't understand what this mean. Is "Vocalizer Embedded" another product that provides the ECI interface? Is it something completely different that voxin happens to know how to drive? Is it available on i386 / x86_64 as well?

gcasse:
Yes, more info at voxin.oralux.net

sthibaul:
> Yes

Yes to what? Both my questions? From what I could find out about Vocalizer, it is available on i386/x86_64, but does not expose the ECI interface.

Which leads me wondering: why not just have a Vocalizer spd module? What benefit do we get from going through a voxin daemon? We already have an `orca -> spd server -> spd module -> pulseaudio` pipeline, I don't see why adding yet another stage, with yet more API translation, between the oldie ECI interface that was not meant for modern use and imposes its own constraints, and the Vocalizer interface which has a proper modern API. Adding another stage also increases synthesis latency.

sthibaul:
As an interim solution, I have merged the voxin module, but factorized with ibmtts to avoid the whole code duplication.

PR #401, opened 2020-09-10
Add missing languages

Nardol:
Since Russian and French have been added recently, these languages was not listed in po/LINGUAS


sthibaul:
Indeed, I didn't realize that weblate was not doing it already. Thanks!

PR #403, opened 2020-09-11
Fix crash with python 3.9

### LINKED ISSUE #402: speech-dispatcher crashes on startup with python 3.9
I noticed that orca is broken in Fedora 33; it starts up "successfully," but never says anything. It's caused by this python crash:

```
12:48:10.345242 - SPEECH: Initializing

Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/orca/speechdispatcherfactory.py", line 161, in __init__
    self._init()
  File "/usr/lib/python3.9/site-packages/orca/speechdispatcherfactory.py", line 172, in _init
    self._client = client = speechd.SSIPClient('Orca', component=self._id)
  File "/usr/lib64/python3.9/site-packages/speechd/client.py", line 578, in __init__
    self._initialize_connection(user, name, component)
  File "/usr/lib64/python3.9/site-packages/speechd/client.py", line 601, in _initialize_connection
    self._conn.send_command('SET', Scope.SELF, 'CLIENT_NAME', full_name)
  File "/usr/lib64/python3.9/site-packages/speechd/client.py", line 326, in send_command
    code, msg, data = self._recv_response()
  File "/usr/lib64/python3.9/site-packages/speechd/client.py", line 292, in _recv_response
    if not self._communication_thread.isAlive():
AttributeError: 'Thread' object has no attribute 'isAlive'

12:48:10.395146 - ERROR: Speech Dispatcher service failed to connect
12:48:10.395666 - SPEECH: Not available
12:48:10.395689 - SPEECH: Initialized
```

[Python 3.9 release notes](https://docs.python.org/3.9/whatsnew/3.9.html) say:

> The isAlive() method of threading.Thread has been removed. It was deprecated since Python 3.8. Use is_alive() instead. (Contributed by Dong-hee Na in bpo-37804.)

Some snarky comment about backwards-compat seems appropriate. Anyway, it looks like Thread.is_alive() has been available since python 2.6, so there should be no risk in changing that.

mcatanzaro:
Thread.isAlive() was removed in python 3.9. Thread.is_alive() has been
available since python 2.6, so let's use that.

Fixes #402

### DIFF
--- a/src/api/python/speechd/client.py
+++ b/src/api/python/speechd/client.py
@@ -289,7 +289,7 @@ class _SSIP_Connection(object):
         and return the triplet (code, msg, data)."""
         # TODO: This check is dumb but seems to work.  The main thread
         # hangs without it, when the Speech Dispatcher connection is lost.
-        if not self._communication_thread.isAlive():
+        if not self._communication_thread.is_alive():
             raise SSIPCommunicationError
         self._ssip_reply_semaphore.acquire()
         # The list is sorted, read the first item


sthibaul:
Thanks!

PR #431, opened 2020-11-13
Mention that spd-say defaults to its locale language

sthibaul:


Inglebard:
Hi,

Should be better placed before "DefaultLanguage" instead "DefaultVoiceType", with the comment "# The Default language with which to speak".

PR #461, opened 2021-01-20
Added a list of terms for gender neutral support in es_ES

lauramcastro:
Namely, pronouns (personal, possessive, demonstrative, undefined), determiners (possessive, demonstrative, some numerals, and indefinite quantifiers).


sthibaul:
Thanks for moving it here :)

Could you rebase your branch over brailcom/speechd's master branch? That'll drop the spurious `Add DefaultVoice option for the generic module` commit which is already in master.

Did you test your changes over actual text?

I realize that in my examples I didn't include upper cases in the `[]`, it should be like this:

`end-words nosotrxs      (?<=\b)[nN][oO][sS][oO][tT][rR][xXeE][sS](?=\b)`

so as to also support upper-case words. You would need to do the same for your rules (as well as update my example, I won't do it myself, to avoid a merge conflict with your changes there).

lauramcastro:
Rebase done.

I would be happy to test the changes over actual text: are there any specific tests in the repo that I should use for that (or as an example to implement some new ones)?

I'll add the uppercase letters as soon as I can.

sthibaul:
> I would be happy to test the changes over actual text: are there any specific tests in the repo that I should use for that (or as an example to implement some new ones)?

We don't have a test infrastructure, but you can test by hand by running spd-say, giving it some spanish text that includes gender-neutral forms.

lauramcastro:
I configured, compiled and installed to a non-default directory. However, running 

`$DESTDIR/usr/local/bin/spd-say -l ES <text in Spanish with gender-neutral words featured in the file gender-neutral.dic>`

does not seem to be different from my system-wide `spd-say`. I guess I must be doing something wrong, could you help me figure out what?

Thanks in advance (and once more) for your time!

sthibaul:
> does not seem to be different from my system-wide spd-say.

You need to make sure that you don't have a speech-dispatcher server already running. You can use

`killall speech-dispatcher`

then

`$DESTDIR/usr/local/bin/speech-dispatcher -t 0 -s`

to make sure that it's your just-installed speech-dispatcher that is getting run.

That being said, beware of using DESTDIR. I believe you rather wanted to give `--prefix` to `./configure`, and then run `make install` without any DESTDIR. That's important, because the binaries you build with look for gender-neutral.dic in `$prefix/share/speech-dispatcher/locale/es/`, and *not* in `$DESTDIR/$prefix/share/speech-dispatcher/locale/es/`.

sthibaul:
Also, that being said again, you don't really need to reconfigure/build/install a speech-dispatcher. Provided that your system has a 0.10 version of speech-dispatcher, you can simply overwrite the `gender-neutral.dic` file in `/usr/local/share/speech-dispatcher/locale/es` and that should just work with your system-provided `spd-say`

lauramcastro:
Yes, I was a bit lazy when going for the `DESTDIR` option rather than re-running `config` with `--prefix`, even though it is not a long process m-).

Anyhow, your second advice makes things much faster to test, and I can acknowledge that it works as expected on the handful of sentences I have tried, where I've made sure every word in gender-neutral.dic is present at least once.

sthibaul:
Ok, I have merged this version, thanks!

Of course, feel free to submit other PR for any further change.

PR #462, opened 2021-02-02
Fix build after glib e38982df

mcatanzaro:
glib.h now includes <type_traits> when building as C++. But kali.cpp
includes it (via module_utils.h) inside an extern "C" block. This is
illegal.

Best practice is to use extern "C" in project header files, not around
include statements. So let's do that. module_utils.h already includes
glib.h, so we can use G_BEGIN_DECLS/G_END_DECLS there. spd_audio.h does
not, so I decided to write out the usual boilerplate, as including all
of glib.h just for G_BEGIN_DECLS/G_END_DECLS seemed like overkill.
Finally, I'll move config.h at the very top of kali.cpp, since I'm
touching this code anyway and that is the usual place to put it.

This is the minimum viable change required for speech-dispatcher to
build with glib master. As long as speech-dispatcher is combining C and
C++, it would be advisible to use extern "C" in all headers that declare
C functions and might be included from C++.

### DIFF
--- a/src/modules/kali.cpp
+++ b/src/modules/kali.cpp
@@ -21,21 +21,21 @@
  * $Id: kali.c,v 1.59 2008-06-09 10:38:02 hanke Exp $
  */
 
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
 #include <stdio.h>
 #include <string.h>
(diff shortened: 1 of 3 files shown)


sthibaul:
Thanks!

PR #551, opened 2021-04-17
Fix htmlpreview doc links in readme

AlyoshaVasilieva:
See https://github.blog/changelog/2021-01-29-github-pages-will-stop-redirecting-pages-sites-from-github-com-after-april-15-2021/


sthibaul:
Thanks!

PR #560, opened 2021-04-19
ivona: Use server-side audio rendering

sthibaul:


sthibaul:
Could somebody test this with a real Ivona server?

PR #603, opened 2021-08-10
increase espeak speed to maximum without sonic

alex19EP:
I'm not sure if I need to change the espeak config I only tested espeak-ng

maybe we should increase the minimum and average speed too.

### DIFF
--- a/config/modules/espeak-ng.conf
+++ b/config/modules/espeak-ng.conf
@@ -33,7 +33,7 @@ EspeakMinRate 80
 EspeakNormalRate 170
 
 # Maximum rate (100 in speech-dispatcher)
-EspeakMaxRate 390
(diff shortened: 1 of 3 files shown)


sthibaul:
It seems espeak as well caps to 450:

```
	if(wpm > 450)
		wpm = 450;
```

so yes we can increase to 449 in both case. Why not setting 450, though?

alex19EP:
> so yes we can increase to 449 in both case. Why not setting 450, though?

because of espeak-ng/espeak-ng#131

sthibaul:
> > so yes we can increase to 449 in both case. Why not setting 450, though?
> 
> because of [espeak-ng/espeak-ng#131](https://github.com/espeak-ng/espeak-ng/issues/131)

ok :)

PR #650, opened 2022-01-08
spd-conf: Improve config copying

jtojnar:
- spd-conf: Do not copy config permission attributes
- spd-conf: Respect XDG basedir spec for autostart

### DIFF
--- a/src/api/python/speechd_config/config.py
+++ b/src/api/python/speechd_config/config.py
@@ -602,7 +602,17 @@ Do you want to keep it?"""), False)
                     return
         # Copy the original intact configuration files
         # creating a conf/ subdirectory
-        shutil.copytree(buildconfig.SPD_CONF_ORIG_PATH, self.test.user_conf_dir())
+        config_root = self.test.user_conf_dir()
+        shutil.copytree(buildconfig.SPD_CONF_ORIG_PATH, config_root)
+        # Ensure the files are writeable when copying from immutable directory.
+        umask = os.umask(0)
+        os.umask(umask)
+        os.chmod(self.test.user_conf_dir() , 0o755 & ~umask)
+        for root, dirs, files in os.walk(self.test.user_conf_dir()):
+            for d in dirs:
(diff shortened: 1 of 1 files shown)


src/api/python/speechd_config/config.py
```diff
@@ -602,7 +602,15 @@ def create_user_configuration(self):
...
-        shutil.copytree(buildconfig.SPD_CONF_ORIG_PATH, self.test.user_conf_dir())
+        config_root = self.test.user_conf_dir()
+        shutil.copytree(buildconfig.SPD_CONF_ORIG_PATH, config_root)
+        # Ensure the files are writeable when copying from immutable directory.
+        os.chmod(self.test.user_conf_dir() , 0o755)
+        for root, dirs, files in os.walk(self.test.user_conf_dir()):
+            for d in dirs:
+                os.chmod(os.path.join(root, d), 0o755)
+            for f in files:
+                os.chmod(os.path.join(root, f), 0o644)
```
sthibaul, line 613: I'd say make sure to `and` with the current umask? Otherwise that could set broader permissions than what the user configured her environment to set by default.

src/api/python/speechd_config/config.py
```diff
@@ -667,8 +675,10 @@ def configure_basic_settings(self, type='user'):
 This is usually not necessary, most applications will start Speech Dispatcher automatically."""),
                 False)
             if setup_autostart:
-                os.system("""cp %s ~/.config/autostart/""" % os.path.join(buildconfig.SPD_DESKTOP_CONF_PATH,
-                                                                          "speechd.desktop"))
+                shutil.copy(
+                    os.path.join(buildconfig.SPD_DESKTOP_CONF_PATH, "speechd.desktop"),
+                    os.path.join(BaseDirectory.xdg_config_home, "autostart")
```
sthibaul, line 680: The ending `/` is important to make sure that we put the file into the `autostart` directory, and not actually create an `autostart` file just because there wasn't any such entry before. So I guess just add `, "/"`

sthibaul:
Thanks!

PR #654, opened 2022-01-09
A more complete .gitignore file

shindere:
This PR gathers the content of all the .gitignore files in a global
one, factorizes patterns and adds more files to ignore.

Two reamining questions:

 1. Rather than doing that, shall we rely on the `git.mk`facility?
 2. Couldn't we stop versionning the HTML manuals so that building speechd keeps git status clean?


sthibaul:
> Rather than doing that, shall we rely on the git.mk facility?

It'd be better to just rely on git.mk,  yes

> Couldn't we stop versionning the HTML manuals so that building speechd keeps git status clean?

We need the html manuals in the git repository for the links in README to work.
We could just drop the "-dirty" part from the version for the documentation generation, though

sthibaul:
I have just dropped the use of `git-version-gen`, it was already giving me headaches, versioning by hand is fine with me :)

sthibaul:
Normally this should be now fixed: .gitignore is just generated, and not commited

PR #751, opened 2022-06-16
Fix pico voice selection

jpwhiting:
I noticed a while ago selecting a voice with spd-say -y when using pico didn't work unless you also specified a language with -l argument.
Today I figured out why. sd_pico first parses the voice name, then parses the language, but parsing the language changes the voice...
Doing those in reverse fixes voice selection to override language selection (which is locale based if none is given to spd-say, etc.)

With this change

spd-say -y virginie "Hello"

uses the virginie voice even though my locale is en-US.

### DIFF
--- a/src/modules/pico.c
+++ b/src/modules/pico.c
@@ -169,7 +169,7 @@ static int pico_process_tts(void)
 
 	text_remaining = strlen((const char *)buf) + 1;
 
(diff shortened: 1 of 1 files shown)


sthibaul:
like is done in other modules indeed, thanks!

PR #752, opened 2022-06-21
Fix libspeechd version.h

jpwhiting:
A previous (post 0.11.1) commit removed defining MAJOR_VERSION, etc. from configure.ac. They are still there as macros, but not exported anymore.

This change adds them back and adds making sure they are defined to clibrary.c test so make check will catch it if it happens again somehow.


sthibaul:
Some version pieces, notably MAJOR_VERSION could be 0 :)
You can simply make the test print them, which will already make sure that they are defined (even if to 0)

jpwhiting:
Yeah, I just noticed that when I ran clibrary test. If I just print them it needs to fail to compile if they aren't defined. Trying, one moment.

sthibaul:
if they aren't defined the code will not compile, that's enough for the test to be effective :)

jpwhiting:
Yep, tested with an #undef in there and compile in make check failed, good. Fixed.

PR #753, opened 2022-06-21
Remove autogenerated documentation files from git.

jpwhiting:
Since these are generated at build time they shouldn't be in git
causing changes when generated, etc.
Having them in git also confuses where fixes should be made.


sthibaul:
I know that poses trouble, but we do want to have the html files available, so they can be linked e.g. from the README.md file, and people find them with google.

Nardol:
As an alternative, could it be possible to create I.E: speechd.github.io sinced from another branch of this repo? Document would still be referenced and not present with the code.

jpwhiting:
I see, ok thanks. Yeah, maybe a statically generated site could be linked to instead later once that gets set up at some point.

sthibaul:
I made `Makefile.am` only regenerate the html files on my system, so that I'll be the only one bothered by the issue :)

(creating a github.io page is not really a simple thing to do...)

PR #754, opened 2022-07-10
spd_wchar: Initialize buffer before using

mgorse:
Wcrtomb doesn't write a null byte after processing the character, but the
buffer is later assumed to be terminated.

### DIFF
--- a/src/api/c/libspeechd.c
+++ b/src/api/c/libspeechd.c
@@ -879,10 +879,10 @@ spd_wchar(SPDConnection * connection, SPDPriority priority, wchar_t wcharacter)
 
 	pthread_mutex_lock(&connection->ssip_mutex);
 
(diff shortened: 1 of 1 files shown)


src/api/c/libspeechd.c
```diff
@@ -879,6 +879,7 @@ spd_wchar(SPDConnection * connection, SPDPriority priority, wchar_t wcharacter)
 
 	pthread_mutex_lock(&connection->ssip_mutex);
 
+	memset(character, 0, sizeof(character));
 	ret = wcrtomb(character, wcharacter, NULL);
 	if (ret <= 0)
 		RET(-1);
```
sthibaul, line 885: I'd say rather just write the ending character `character[ret] = '\0';` ?

sthibaul:
Thanks!

PR #755, opened 2022-07-11
Add missing pthread include

Redfoxymoon:
ditto. I don't understand how this has worked before?


sthibaul:
glibc's stdlib.h seems to be pulling pthread types in, that's probably why :)

sthibaul:
Note however that that `pthread_main.h` is quite recent.

Issue #611, opened 2021-09-15
speech-dispatcher fails to build with slibtool (undefined reference to `lt__PROGRAM__LTX_preloaded_symbols`)


CyberTailor:
It builds with GNU libtool but not with any of slibtool variants.

Relevant fragment:
```
rlibtool: link: x86_64-pc-linux-gnu-gcc dummy.o spd_audio.o module_main.o module_utils.o module_utils_addvoice.o ../../src/common/.libs/libcommon.a -Wall -march=bdver2 -O2 -pipe -fpch-preprocess -ggdb -frecord-gcc-switches -Wl,-O1 -Wl,--as-needed -Wl,--defsym=__gentoo_check_ldflags__=0 -Wl,-O1 -Wl,--as-needed -Wl,--defsym=__gentoo_check_ldflags__=0 -lglib-2.0 -lc -lltdl -lpthread -lm -lsndfile -ldotconf -lglib-2.0 -lgthread-2.0 -pthread -lglib-2.0 -lc -lltdl -lpthread -lm -o .libs/sd_dummy
/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: module_main.o: warning: relocation against `lt__PROGRAM__LTX_preloaded_symbols' in read-only section `.text.startup'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: module_main.o: in function `main':
/var/tmp/portage/app-accessibility/speech-dispatcher-0.9.1/work/speech-dispatcher-0.9.1/src/modules/module_main.c:79: undefined reference to `lt__PROGRAM__LTX_preloaded_symbols'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: warning: creating DT_TEXTREL in a PIE
collect2: error: ld returned 1 exit status
rlibtool: exec error upon slbt_exec_link_create_executable(), line 1745: (see child process error messages).
rlibtool: < returned to > slbt_exec_link(), line 2155.
make[3]: *** [Makefile:805: sd_dummy] Error 2
make[3]: Leaving directory '/var/tmp/portage/app-accessibility/speech-dispatcher-0.9.1/work/speech-dispatcher-0.9.1/src/modules'
make[2]: *** [Makefile:453: all-recursive] Error 1
[...]

PR #756, opened 2022-07-11
drop ltdl dependency in favor of native dlopen(3) interfaces

Redfoxymoon:
there's no platform(s) where ltdl is necessary anymore that anyone would want to build and use speechd on, and this fixes building speechd w/ slibtool. #611 

however I don't know how to properly test the module loading to make sure it's correct so please don't merge before it's verified to actually work... :-)

### DIFF
--- a/src/common/spd_audio.c
+++ b/src/common/spd_audio.c
@@ -47,9 +47,16 @@
 #include <pthread.h>
 
 #include <glib.h>
+#ifdef USE_DLOPEN
+#include <dlfcn.h>
+#else
 #include <ltdl.h>
+#endif
 
 static int spd_audio_log_level;
+#ifdef USE_DLOPEN
+static void *dlhandle;
+#else
 static lt_dlhandle lt_h;
 
 /* Dynamically load a library with RTLD_GLOBAL set.
@@ -72,6 +79,7 @@ static lt_dlhandle my_dlopenextglobal(const char *filename)
 	lt_dladvise_destroy(&advise);
 	return handle;
 }
+#endif
 
 /* Open the audio device.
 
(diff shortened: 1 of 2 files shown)


Redfoxymoon:
Right, it should build on older glibc boxes now, silly libdl detection...

sthibaul:
> there's no platform(s) where ltdl is necessary anymore that anyone would want to build and use speechd on

We never really know.

Can't we rather keep the ldtl code, and have the dlopen code as an alternative when libltdl is not available?

Redfoxymoon:
That can be done yes

Redfoxymoon:
I've updated the PR to have two separate codepaths for ltdl and dlopen(3), I've left out a potential configure flag however, not sure how that would be best handled...

sthibaul:
> I've updated the PR to have two separate codepaths for ltdl and dlopen(3)

That looks good, thanks!

>  I've left out a potential configure flag however, not sure how that would be best handled...

I'd say change the 

```
AC_SEARCH_LIBS([lt_dlopen], [ltdl], [],
	[AC_MSG_FAILURE([ltdl library missing])])
```

test: when it fails, instead of an error, enable using `dlopen(3)`

Redfoxymoon:
the primary use-case for the dlopen(3) codepath is linux, where ltdl is overwhelmingly common still so that would be kind-of backwards

sthibaul:
> the primary use-case for the dlopen(3) codepath is linux, where ltdl is overwhelmingly common still so that would be kind-of backwards

? I don't understand what you mean. You are here saying that ltdl is common, but at the same time your PR is aimed at being able to avoid to depend on it? If ltdl is that common, then we can depend on it. If it's not that common, then it makes sense to make the ltdl probing failure non-fatal by adding the dlopen case.

Redfoxymoon:
GNU libtool and libltdl in turn are basically un-maintained at this point which isn't very good for new unix-likes that may want to use speechd, there's also a replacement to GNU libtool called [https://dev.midipix.org/cross/slibtool](slibtool) which currently lacks support for ltdl

sthibaul:
Ok, sure, but how is it related? Really I don't understand the ins and outs of what you are saying.

What is the *actual* issue that is left with what I propose?

Is it that slibtool is wrongly taken as a libtool-compatible replacement?

Then it's the test that needs to be fixed. From #611 it seems that it's `LTDL_SET_PRELOADED_SYMBOLS` which is missing. Then make configure.ac check for that, and only use ltdl if that successes. That way, dlopen will be used whenever the real libtool is not available.

PR #760, opened 2022-08-25
Allow building without ltdl

sthibaul:


orbea:
@sthibaul Thanks a lot for doing this!

Ideally slibtool would just work with ltdl, but its easier said than done and there are very few packages in gentoo that depend on it.

PR #776, opened 2022-09-30
Only install user units if libsystemd is available

nteodosio:
This is a follow up to merge request #763.

If Libsystemd were not available to the system, the previous implementation would nonetheless install both the socket and the dispatcher.service, which, although not a bug, wouldn't be used.

I'm also renaming HAVE_SYSTEMD to USE_SYSTEMDLIB because 

1. that's the module being checked
2. even if available, the user could choose not to use it by passing `--without-systemduserunitdir`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -118,11 +118,7 @@ PKG_CHECK_MODULES([SNDFILE], [sndfile >= 1.0.2])
 AC_SUBST([SNDFILE_CFLAGS])
 AC_SUBST([SNDFILE_LIBS])
 
-PKG_CHECK_MODULES([LIBSYSTEMD], [libsystemd], [have_systemd="yes"], [have_systemd="no"])
-if test x$have_systemd = xyes; then
-    AC_DEFINE([HAVE_SYSTEMD], [1], [Define to 1 if you have the systemd library])
-fi
-AM_CONDITIONAL(HAVE_SYSTEMD, [test x$have_systemd = xyes])
+PKG_CHECK_MODULES([LIBSYSTEMD], [libsystemd], [have_libsystemd=yes], [:])
 
 # Checks for header files.
 AC_CHECK_HEADERS([arpa/inet.h fcntl.h langinfo.h limits.h netdb.h])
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -551,10 +547,11 @@ AM_CONDITIONAL(HAVE_SYSTEMD, [test -n "$with_systemdsystemunitdir" -a "x$with_sy
 AC_ARG_WITH([systemduserunitdir],
         AS_HELP_STRING([--with-systemduserunitdir=DIR], [Directory for systemd user service files]),
         [], [with_systemduserunitdir=$($PKG_CONFIG --variable=systemduserunitdir systemd)])
-if test "x$with_systemduserunitdir" != xno; then
+AM_CONDITIONAL([USE_LIBSYSTEMD], [test -n "$have_libsystemd" -a -n "$with_systemduserunitdir" -a "x$with_systemduserunitdir" != xno])
+if test -n "$have_libsystemd" -a -n "$with_systemduserunitdir" -a "x$with_systemduserunitdir" != xno ; then
         AC_SUBST([systemduserunitdir], [$with_systemduserunitdir])
+        AC_DEFINE([USE_LIBSYSTEMD], [1], [Used for activation via socket])
```
sthibaul, line 553: Rather call AM_CONDITIONAL after the if, using the USE_LIBSYSTEMD variable in the test. Otherwise sooner or later, somebody will modify line 550 without thinking about modifying line 551, and we'll get incoherent

sthibaul:
Right, that makes sense :)

configure.ac
nteodosio: I should have mentioned that I tried that but autoconf complains.

>  Note that you must arrange for every AM_CONDITIONAL to be invoked every time configure is run. If AM_CONDITIONAL is run conditionally (e.g., in a shell if statement), then the result will confuse automake. 

https://www.gnu.org/software/automake/manual/html_node/Usage-of-Conditionals.html

Not sure what is the best way to circumvent that?

sthibaul: I'm not saying to put it inside the if, but *after* it, and use the USE_LIBSYSTEMD variable to know whether the if succeeded or not.

nteodosio: Please don't merge yet, I must have be misunderstanding, because
```
AM_CONDITIONAL([USE_LIBSYSTEMD], [USE_LIBSYSTEMD])
```
wouldn't work.

sthibaul: See how it's done with other, here it's be `[test x$USE_LIBSYSTEMD = x1]`

sthibaul: Oh, no, sorry, AC_DEFINE doesn't actually define an environment variable. Then define one.

nteodosio: Thanks for your help! I tested and works as intended with/without `libsystemd-dev` on the host.

sthibaul:
Thanks!

PR #778, opened 2022-10-07
libspeechd: Avoid crashing if the connection got broken

### LINKED ISSUE #777: spd-day displays invalid language error and segfault
Using Ubuntu Kinetic

ii  libspeechd2:amd64                          0.11.3-1ubuntu1                            amd64        Speech Dispatcher: Shared libraries
ii  python3-speechd                            0.11.3-1ubuntu1                            all          Python interface to Speech Dispatcher
ii  speech-dispatcher                          0.11.3-1ubuntu1                            amd64        Common interface to speech synthesizers
ii  speech-dispatcher-audio-plugins:amd64      0.11.3-1ubuntu1                            amd64        Speech Dispatcher: Audio output plugins
ii  speech-dispatcher-espeak-ng                0.11.3-1ubuntu1                            amd64        Speech Dispatcher: Espeak-ng output module

That's basically 0.11.3-1 with the socket activation change from !763 cherrypicked
http://launchpadlibrarian.net/626414937/speech-dispatcher_0.11.3-1_0.11.3-1ubuntu1.diff.gz

```
$ spd-say -L
Invalid language fr-FR!
Erreur de segmentation (core dumped)

#0  __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:76
No locals.
#1  0x00007ffff7f92bda in get_param_str (reply=reply@entry=0x0, num=num@entry=1, err=err@entry=0x7fffffffdb3c)
    at /tmp/build/speech-dispatcher-0.11.3/src/api/c/libspeechd.c:2003
        i = <optimized out>
        tptr = 0x7fffffffdc04 ""
        pos = 0x0
        pos_begin = <optimized out>
        pos_end = <optimized out>
        rep = <optimized out>
        __PRETTY_FUNCTION__ = <optimized out>
#2  0x00007ffff7f939a7 in spd_execute_command_with_list_reply (connection=connection@entry=0x555555567980, command=command@entry=0x7ffff7f970c7 "LIST SYNTHESIS_VOICES")
    at /tmp/build/speech-dispatcher-0.11.3/src/api/c/libspeechd.c:1541
        reply = 0x0
        line = <optimized out>
        err = 21845
        max_items = 50
        result = 0x555555567d20
        i = 0
#3  0x00007ffff7f93a6e in spd_list_synthesis_voices (connection=connection@entry=0x555555567980) at /tmp/build/speech-dispatcher-0.11.3/src/api/c/libspeechd.c:1476
        svoices_str = <optimized out>
        svoices = <optimized out>
        i = <optimized out>
        num_items = <optimized out>
        __PRETTY_FUNCTION__ = "spd_list_synthesis_voices"
#4  0x000055555555718b in main (argc=2, argv=0x7fffffffdd68) at ./src/clients/say/say.c:199
        list = <optimized out>
        i = <optimized out>
        conn = 0x555555567980
        spd_priority = <optimized out>
        err = <optimized out>
        error = 0x7ffff7ca06a2 <__GI___libc_malloc+418> "H\205\300\017\204\211"
        msg_arg_required = <optimized out>
        ret = <optimized out>
        option_ret = <optimized out>
        line = <optimized out>
```

the issue seems random, sometime it returns an empty list and sometime the expected result. The previous revision without the systemd change doesn't seem to hit the problem so it could be a regression (or an existing issue made more easy to hit for $reasons)

Cc @nteodosio

sthibaul:
during spd_execute_command_with_reply

Fixes #777

### DIFF
--- a/src/api/c/libspeechd.c
+++ b/src/api/c/libspeechd.c
@@ -1529,7 +1529,7 @@ char **spd_execute_command_with_list_reply(SPDConnection * connection,
 	int i;
 
 	spd_execute_command_with_reply(connection, command, &reply);
-	if (!ret_ok(reply)) {
+	if (ret_ok(reply) <= 0) {
 		if (reply != NULL)
 			free(reply);
 		return NULL;


nteodosio:
### With disabled socket: `systemctl stop --user speech-dispatcher.socket`

`spd-say` still occasionally fails with the language error. So I couldn't observe a change in behavior.

### With disabled socket: `systemctl start --user speech-dispatcher.socket`

This change makes the flow to reach inside the `if (sd_listen_fds(0) >= 1)`, which earlier today — i.e., without the change proposed here — I realized wouldn't be reached! 

Nonetheless, either no sound is emitted and spd-say exits with code 141 (SIGPIPE!) or it fails with the same invalid language message.

sthibaul:
> Nonetheless, either no sound is emitted and spd-say exits with code 141 (SIGPIPE!) or it fails with the same invalid language message.

Ok, but does the change avoid the segfault message? Which means it'd the right thing we are after.

nteodosio:
I can only get the segfault message with the pkill based test case, which, as I gather from your comment on #777, is not ideal. 

Otherwise no, I don't get the segfault message.

nteodosio:
If the Systemd socket is enabled, however, I do either get 141 or the segfault.

```
% spd-say hi  #Exits with 141
% spd-say hi
Invalid language en-US!
Speech Dispatcher failed to say message
```

sthibaul:
@nteodosio https://github.com/brailcom/speechd/pull/817 probably fixed that issue?

sthibaul:
That was commited as 4a24a12929a2b0566435e06cce3565c126bfc58f

nteodosio:
Yes, indeed, thank you!

Issue #782, opened 2022-10-17
Module loading does not work with dlopen (--disable-ltdl)


voyageur:
I found this on Gentoo Linux, where we use the new --disable-ltdl flag on recent 0.11.3 (and 0.11.2) which includes #756.

Trying to load modules with dlopen fails on my system, tested with a sd_generic module (piping in gTTS) and dummy module.

Configure command:
`./configure --prefix=/usr --build=x86_64-pc-linux-gnu --host=x86_64-pc-linux-gnu --mandir=/usr/share/man --infodir=/usr/share/info --datadir=/usr/share --sysconfdir=/etc --localstatedir=/var/lib --datarootdir=/usr/share --disable-dependency-tracking --disable-silent-rules --disable-static --docdir=/usr/share/doc/speech-dispatcher-0.11.3 --htmldir=/usr/share/doc/speech-dispatcher-0.11.3/html --with-sysroot=/ --libdir=/usr/lib64 --disable-ltdl --disable-python --disable-static --with-baratinoo=no --with-ibmtts=no --with-kali=no --with-pico=no --with-voxin=no --with-alsa --without-libao --without-espeak --without-espeak-ng --without-flite --with-nas --with-pulse --with-systemdsystemunitdir=/lib/systemd/system`

First issue is dlopen command does not find the library:
```
==> /tmp/speechd-debug/dummy.log <==
 Mon Oct 17 11:41:26 2022 [20633]: Additional logging initialized
 Mon Oct 17 11:41:26 2022 [20920]: Opening audio output system
 Mon Oct 17 11:41:26 2022 [20964]: Can't use server: Cannot open plugin server. error: /usr/lib64/speech-dispatcher/spd_server: cannot open shared object file: No such file or directory
 Mon Oct 17 11:41:26 2022 [21191]: Opening audio output system
 Mon Oct 17 11:41:26 2022 [21208]: Can't use pulse: Cannot open plugin pulse. error: /usr/lib64/speech-dispatcher/spd_pulse: cannot open shared object file: No such file or directory

==> /tmp/speechd-debug/gtts-generic.log <==
 Mon Oct 17 11:41:26 2022 [22403]: Additional logging initialized
 Mon Oct 17 11:41:26 2022 [22536]: Opening audio output system
 Mon Oct 17 11:41:26 2022 [22565]: Can't use server: Cannot open plugin server. error: /usr/lib64/speech-dispatcher/spd_server: cannot open shared object file: No such file or directory
[...]

PR #783, opened 2022-10-17
Fix audio plugin loading with dlopen #782

voyageur:
dlopen needs the full name so include .so extension
Set SPD_AUDIO_PLUGIN_ENTRY to correct name (ltdl adds a prefix, dlopen just uses standard name)


CyberGuerro:
Is it possible to have a new release so I can create a new installation package for my distribution (Gentoo)?

regards

PR #794, opened 2022-11-23
espeak: set buffer size to 300

### LINKED ISSUE #793: espeak-ng: strange sounds when reading some strings
I am 90% sure that this is another bug in espeak's ssml but since I could not repeat it through espeak-ng cli but only through spd-say, I decided to write here first.

### Steps to reproduce

`spd-say -o espeak-ng -l 'ru' -y 'Russian+Max' -r 100 -w -x "<speak><mark name="0:4"/>read <mark name="5:10"/>bytes <mark name="11:15"/>from <mark name="16:21"/>their <mark name="22:27"/>input <mark name="28:31"/>and <mark name="32:37"/>write <mark name="38:41"/>the <mark name="42:47"/>bytes <mark name="48:50"/>to <mark name="51:56"/>their <mark name="57:64"/>output.<mark name="__spd_0"/> <mark name="65:68"/>The <mark name="69:73"/>real <mark name="74:79"/>magic <mark name="80:82"/>is <mark name="83:85"/>in <mark name="86:95"/> косая черта dev косая черта sr0.<mark name="__spd_1"/></speak>"`

### Obtained behavior

listen for a strange gurgling overtone on the second bytes word.

### Expected behavior

sound should be same as when executing:
`espeak-ng -v ru+max -s 449 -m '<speak><mark name="0:4"/>read <mark name="5:10"/>bytes <mark name="11:15"/>from <mark name="16:21"/>their <mark name="22:27"/>input <mark name="28:31"/>and <mark name="32:37"/>write <mark name="38:41"/>the <mark name="42:47"/>bytes <mark name="48:50"/>to <mark name="51:56"/>their <mark name="57:64"/>output.<mark name="__spd_0"/> <mark name="65:68"/>The <mark name="69:73"/>real <mark name="74:79"/>magic <mark name="80:82"/>is <mark name="83:85"/>in <mark name="86:95"/> косая черта dev косая черта sr0.<mark name="__spd_1"/></speak>'`

### Distribution

Arch Linux

### Version of Speech-dispatcher

0.11.3.r17.ga3acf81a
espeak-ng: 1.51.r587.g0f3fb42b

alex19EP:
this fixes #793
ref for buffer size: https://github.com/nvaccess/nvda/blob/a6fb2392083b5fa1bae926102135ad452746ad3c/source/synthDrivers/_espeak.py#L338

### DIFF
--- a/config/modules/espeak-ng-mbrola.conf
+++ b/config/modules/espeak-ng-mbrola.conf
@@ -41,7 +41,7 @@ EspeakMaxRate 449
 # -- Internal parameters --
 
 # Number of ms of audio returned by the espeak callback function.
-EspeakAudioChunkSize 3000
+EspeakAudioChunkSize 300
 
 # Maximum number of samples to buffer in playback queue.
 EspeakAudioQueueMaxSize 441000
--- a/config/modules/espeak-ng.conf
+++ b/config/modules/espeak-ng.conf
@@ -41,7 +41,7 @@ EspeakMaxRate 449
 # -- Internal parameters --
 
 # Number of ms of audio returned by the espeak callback function.
-EspeakAudioChunkSize 3000
+EspeakAudioChunkSize 300
 
 # Maximum number of samples to buffer in playback queue.
 EspeakAudioQueueMaxSize 441000
--- a/config/modules/espeak.conf
+++ b/config/modules/espeak.conf
@@ -41,7 +41,7 @@ EspeakMaxRate 449
 # -- Internal parameters --
 
 # Number of ms of audio returned by the espeak callback function.
-EspeakAudioChunkSize 3000
+EspeakAudioChunkSize 300
 
 # Maximum number of samples to buffer in playback queue.
 EspeakAudioQueueMaxSize 441000


sthibaul:
There is really no reason why we should reduce a buffer size in order to work around synthesis bugs. That would just *hide* the bug and let it have other less-visible consequences.

alex19EP:
> There is really no reason why we should reduce a buffer size in order to work around synthesis bugs. That would just hide the bug and let it have other less-visible consequences.

There is really no reason why we should torment Linux users with annoying bug even longer.
or force to edit the configuration files to fix it.

moreover I don't see the point of setting so large buffer. NVDA uses 300 for years and users are happy.

please reconsider

sthibaul:
> and users are happy.

Except that the bug probably also appears in obscure ways that are harder to debug.

Really, let's just fix bugs, not work around them and "happy" with that.

djphoenix:
@sthibaul I really understand your position as project maintainer. However, unfortunately, a phrase “just fix a bug” seems not reliable when we have a 7-10 years old legacy.

The buffer size affects many, many routines in espeak, and no one understand how for now. I really want to fix it, but have no estimates.

Yes, buffer reduction is a workaround (crutch, if you want to hear it). But hey, crutches helps people to walk!

sthibaul:
> I really want to fix it

Ok, that's what I wanted to hear.

PR #808, opened 2023-01-26
python: Fix checking all lines of a message have the same code

cwendling:
This is something that was lost back in 2003 in commit c71d08a1de6888f8be6a007303d18bd8bcce8c65.

This is merely a missing validation, but the code looks very strange as it was, trying to perform the validation but having it not actually do anything.

---

Disclaimer: I didn't actually test this very much, and possibly the extra validation would show bugs elsewhere.  However, I do think something should be fixed in this area: either have fully working code, or remove the pieces that pretend to check yet are no-ops.


sthibaul:
That's right indeed, thanks!

PR #825, opened 2023-04-12
Update CLDR to version 43 and NVDA symbols to latest version from source

Nardol:
CLDR version 43 has been published on unicode.org.
First attempt from git source added some errors in the dictionaries which are now fixed using zip file from the website.
It confirms it is better not to use the Git repository, continuing as done until now is the best way.


locale/Makefile.am
```diff
@@ -30,6 +30,7 @@ nobase_localedata_DATA = \
 	az/emojis.dic \
 	be/emojis.dic \
 	bg/emojis.dic \
+	bgn/emojis.dic
```
sthibaul, line 33: These are missing a trailing backslash

sthibaul (changes requested):

locale/Makefile.am
Nardol: Oops, fixed

PR #832, opened 2023-06-02
Speed up spd_execute_command_with_list_reply

novalis:
spd_execute_command_with_list_reply was accidentally quadratic: it would start by searching the string for one \r\n, then it would search for two, then for three...  I have 15,000 voices (I have no idea why; it seems to be the default Debian config), so this was quite slow.

For me, this improves the performance of spd_list_synthesis_voices from 1.5s to 0.5s.

### DIFF
--- a/src/api/c/libspeechd.c
+++ b/src/api/c/libspeechd.c
@@ -70,6 +70,7 @@ static int isanum(char *str);
 static char *get_reply(SPDConnection * connection);
 static int get_err_code(char *reply);
 static char *get_param_str(char *reply, int num, int *err);
+static char *get_param_str_and_advance(char **reply, int *err);
 static int get_param_int(char *reply, int num, int *err);
 static int ret_ok(char *reply);
 static void SPD_DBG(char *format, ...);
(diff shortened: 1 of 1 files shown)


sthibaul:
Thanks for your patch, it looks promising :)

I'd however rather not copy/paste the implementation of `get_param_str`.

It seems to me like everything in `get_param_str` after the for loop could be replaced by a call to your function, thus factorizing everything that can be?

src/api/c/libspeechd.c
```diff
@@ -542,7 +543,7 @@ SPDConnection *spd_open2(const char *client_name, const char *connection_name,
 	if (!connection->stream)
 		SPD_FATAL("Can't create a stream for socket, fdopen() failed.");
 	/* Switch to line buffering mode */
-	ret = setvbuf(connection->stream, NULL, _IONBF, SPD_REPLY_BUF_SIZE);
+	ret = setvbuf(connection->stream, NULL, _IOLBF, SPD_REPLY_BUF_SIZE);
```
sthibaul, line 546: Please do not mix with other fixes :)

sthibaul (changes requested):

src/api/c/libspeechd.c
novalis: oops

src/api/c/libspeechd.c
```diff
@@ -2020,17 +2008,44 @@ static char *get_param_str(char *reply, int num, int *err)
 
 	pos_begin = pos + 4;
 	pos_end = strstr(pos_begin, "\r\n");
-	if (pos_end == NULL) {
+	if (*pos_end == 0) {
```
sthibaul, line 2011: Noticing this just now: there shouldn't be a `*`, it's really the pointer that needs to be tested. You can leave the code as it is.

sthibaul (approved):

sthibaul:
Thanks!

PR #837, opened 2023-06-06
client library: fix buffering

### LINKED ISSUE #750: speechd doesn't work correctly on Linux using musl libc
### Steps to reproduce
The easiest way for me to reproduce this behaviour is to build the KDE package `pimcommon` and run its test suite, specifically `translatortest`, but any application that uses speechd's API with threads will eventually hit this issue.

### Obtained behavior
The function call `spd_close` never returns because the `spd_events_handler` thread never returns from `getline(3)`.

### Expected behavior
The process should quit when it exits.

### Behavior information
This is because `getline(3)` is not a cancellation point in musl.

```
(gdb) run
Starting program: /home/awilcox/Code/awilfox/user-next/user/pimcommon/src/pimcommon-22.04.2/bin/translatortest 
********* Start testing of TranslatorTest *********
Config: Using QtTest library 5.15.4, Qt 5.15.4 (power64-big_endian-lp64 shared (dynamic) release build; by GCC 8.5.0), adelie 1.0-rc2
PASS   : TranslatorTest::initTestCase()
[New LWP 47005]
[New LWP 47006]
[Detaching after fork from child process 47007]
[New LWP 47029]
PASS   : TranslatorTest::shouldHaveDefaultValuesOnCreation()
PASS   : TranslatorTest::shouldEnableTranslateButtonWhenTextToTranslateIsNotEmpty()
PASS   : TranslatorTest::shouldDisableTranslateButtonAndClearTextWhenClickOnClearButton()
PASS   : TranslatorTest::shouldInvertLanguageWhenClickOnInvertButton()
QWARN  : TranslatorTest::shouldHideWidgetWhenPressEscape() This plugin does not support propagateSizeHints()
PASS   : TranslatorTest::shouldHideWidgetWhenPressEscape()
QWARN  : TranslatorTest::shouldEmitTranslatorWasClosedSignalWhenCloseIt() This plugin does not support propagateSizeHints()
PASS   : TranslatorTest::shouldEmitTranslatorWasClosedSignalWhenCloseIt()
PASS   : TranslatorTest::cleanupTestCase()
Totals: 8 passed, 0 failed, 0 skipped, 0 blacklisted, 324ms
********* Finished testing of TranslatorTest *********
[LWP 47005 exited]

Thread 4 "translatortest" received signal ?, Unknown signal.
[Switching to LWP 47029]
0x00003ffff7f81acc in __syscall3 (c=2, b=70368468612984, a=6, n=145) at ./arch/powerpc64/syscall_arch.h:41
41	./arch/powerpc64/syscall_arch.h: No such file or directory.
(gdb) bt
#0  0x00003ffff7f81acc in __syscall3 (c=2, b=70368468612984, a=6, n=145) at ./arch/powerpc64/syscall_arch.h:41
#1  __stdio_read (f=0x3fffef976aa0, buf=0x3fffef9337e7 " \035", len=1) at src/stdio/__stdio_read.c:12
#2  0x00003ffff7f82050 in __uflow (f=0x3fffef976aa0) at src/stdio/__uflow.c:9
#3  0x00003ffff7f867cc in getdelim (s=0x3fffef9338f8, n=0x3fffef9338e8, delim=<optimized out>, f=0x3fffef976aa0)
    at src/stdio/getdelim.c:64
#4  0x00003ffff7f86b38 in getline (s=<optimized out>, n=<optimized out>, f=<optimized out>)
    at src/stdio/getline.c:5
#5  0x00003fffef937504 in get_reply (connection=connection@entry=0x3fffef978e80) at libspeechd.c:1796
#6  0x00003fffef937ad8 in spd_events_handler (conn=0x3fffef978e80) at libspeechd.c:1838
#7  0x00003ffff7f9ba98 in start (p=0x3fffef933a20) at src/thread/pthread_create.c:203
#8  0x00003ffff7fb083c in __clone () at src/thread/powerpc64/clone.s:44
(gdb) c
Continuing.
warning: Signal ? does not exist on this system.
```

### Distribution
Adélie Linux on 64-bit PowerPC, though it has also been duplicated on x86_64.

### Version of Speech-dispatcher
0.11.1.

### LINKED ISSUE #833: line buffering vs character buffering
### Steps to reproduce

As I noted in #832, I have about 15,000 voices (I don't know why; I think I've just got default Debian).  So spd_list_synthesis_voices is slow, taking (after that PR) about 0.5s.  I can reduce that to 0.35s with:

```
diff --git a/src/api/c/libspeechd.c b/src/api/c/libspeechd.c
index 2ef3b9ea..c7b54feb 100644
--- a/src/api/c/libspeechd.c
+++ b/src/api/c/libspeechd.c
@@ -543,7 +543,7 @@ SPDConnection *spd_open2(const char *client_name, const char *connection_name,
        if (!connection->stream)
                SPD_FATAL("Can't create a stream for socket, fdopen() failed.");
        /* Switch to line buffering mode */
-       ret = setvbuf(connection->stream, NULL, _IONBF, SPD_REPLY_BUF_SIZE);
+       ret = setvbuf(connection->stream, NULL, _IOLBF, SPD_REPLY_BUF_SIZE);
        if (ret)
                SPD_FATAL("Can't set buffering, setvbuf failed.");
``` 

Note that this was the original code, before ec25ad9ada11327a6b890a4bf5b4036fa7e1d472 (in 2005).  Notice also the comment claiming this sets line buffering mode: while the code was changed, the comment was not.   

### Obtained behavior

`spd-say -L` reads one character at a time, leading to over 400k `read` syscalls.

### Expected behavior

Looks like 208 `read` syscalls afterwards (although some may be from other sources; at any rate, client i/o is no longer the bottleneck).

Note that https://github.com/brailcom/speechd/blob/95e7a9cb1f1e16a76857f525e31ec3b201e2ab2b/src/server/module.c#L502 has the same problem (including bogus comment), dating from  541f955205f0020375ec293e584268fdbbb5ee3f (committed a few minutes after ec25ad9ada11327a6b890a4bf5b4036fa7e1d472)

I didn't try changing that to see if it would further improve perf, since that seemed more complicated to test (since I am already running my distro's version of speech-dispatcher).  But I bet it would.

### Distribution

Debian

### Version of Speech-dispatcher

My comments are as-of 95e7a9cb1f1e16a76857f525e31ec3b201e2ab2b

novalis:
In ec25ad9, buffering was switched from line buffering to character buffering.  This caused slow performance when large amounts of data are transferred. Instead, make each connection maintain a buffer of size SPD_REPLY_BUF_SIZE.

For me, this improves the performance of `spd-say -L > /dev/null` from 0.5s to 0.35s.

This might also fix #750, or perhaps we will also need to fix the server side as well.

Fixes #833

### DIFF
--- a/src/api/c/libspeechd.c
+++ b/src/api/c/libspeechd.c
@@ -89,8 +89,8 @@ struct SPDConnection_threaddata {
 };
 
 /*
- * Added by Willie Walker - strndup and getline were GNU libc extensions
- * that were adopted in the POSIX.1-2008 standard, but are not yet found
+ * Added by Willie Walker - strndup was a GNU libc extensions
+ * that was adopted in the POSIX.1-2008 standard, but is not yet found
  * on all systems.
  */
 #ifndef HAVE_STRNDUP
@@ -114,52 +114,64 @@ char *strndup(const char *s, size_t n)
 }
 #endif /* HAVE_STRNDUP */
 
-#ifndef HAVE_GETLINE
-#define BUFFER_LEN 256
-ssize_t getline(char **lineptr, size_t * n, FILE * f)
-{
-	int ch;
-	size_t m = 0;
-	ssize_t buf_len = 0;
-	char *buf = NULL;
-	char *p = NULL;
+#define SPD_REPLY_BUF_SIZE 65536
 
-	if (errno != 0) {
-		SPD_DBG("getline: errno came in as %d!!!\n", errno);
-		errno = 0;
-	}
-	while ((ch = getc(f)) != EOF) {
-		if (errno != 0)
-			return -1;
-		if (m++ >= buf_len) {
-			buf_len += BUFFER_LEN;
-			buf = (char *)realloc(buf, buf_len + 1);
-			if (buf == NULL) {
-				SPD_DBG("buf==NULL");
-				return -1;
+/** Read from the connection's buffer or socket until a newline is
+    read.  Return a pointer to the beginning of the data, and the
+    length of the data (including newline).  The returned value will
+    not be null-terminated, and will include the newline.  Note that
+    the returned pointer will only be valid until the next call to
+    get_line.
+
+    If, after SPD_REPLY_BUF_SIZE bytes, there is no newline (or in the
+    event of a read error) return NULL.
+
+    Unlike getline, this does not handle embedded \0 bytes.
+*/
+char *get_line(SPDConnection * conn, int *n)
+{
+	int bytes;
+	int i;
+	char *ret = NULL;
+	int search_start = conn->buf_start;
+	int message_prefix_len;
+
+	while (1) {
+		for (i = search_start; i < conn->buf_used; i++) {
+			if (conn->buf[i] == '\n') {
+				*n = i + 1 - conn->buf_start;
+				ret = conn->buf + conn->buf_start;
+				conn->buf_start = i + 1;
+				return ret;
 			}
-			p = buf + buf_len - BUFFER_LEN;
 		}
-		*p = ch;
-		p++;
-		if (ch == '\n')
-			break;
+
+		if (conn->buf_start != 0) {
+			message_prefix_len = conn->buf_used - conn->buf_start;
+			memmove(conn->buf, conn->buf + conn->buf_start,
+				message_prefix_len);
+			search_start = message_prefix_len;
+			conn->buf_used = message_prefix_len;
+			conn->buf_start = 0;
+		}
+
+		if (conn->buf_used == SPD_REPLY_BUF_SIZE) {
+			SPD_FATAL
+			    ("No newline after reading SPD_REPLY_BUF_SIZE");
+			return NULL;
+		}
+
+		bytes =
+		    read(conn->socket, conn->buf + conn->buf_used,
+			 SPD_REPLY_BUF_SIZE - conn->buf_used);
+		if (bytes == -1)
+			return NULL;
+		conn->buf_used += bytes;
 	}
-	if (m == 0) {
-		SPD_DBG("getline: m=%d!", m);
-		return -1;
-	} else {
-		*p = '\0';
-		*lineptr = buf;
-		*n = m;
-		return m;
-	}
+
 }
-#endif /* HAVE_GETLINE */
 
 /* --------------------- Public functions ------------------------- */
-
-#define SPD_REPLY_BUF_SIZE 65536
 
 /* Determine address for the unix socket */
 static char *_get_default_unix_socket_name(void)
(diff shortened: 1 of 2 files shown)


src/api/c/libspeechd.c
```diff
@@ -114,53 +114,65 @@ char *strndup(const char *s, size_t n)
...
+
+/** Read from the connection's buffer or socket until a newline is
+    read.  Return a pointer to the beginning of the data, and the
+    length of the data (including newline).  The returned value will
+    not be null-terminated, and will include the newline.  Note that
+    the returned pointer will only be valid until the next call to
+    get_line.
+
+    If, after SPD_REPLY_BUF_SIZE bytes, or in the event of a read
+    error, there is no newline, return NULL.
```
sthibaul, line 127: "there is no newline" should be before  the "or in the event", otherwise the sentence looks odd :)

src/api/c/libspeechd.c
```diff
@@ -562,7 +580,7 @@ SPDConnection *spd_open2(const char *client_name, const char *connection_name,
 		if (ret != 0) {
 			*error_result = strdup("Thread initialization failed");
 			SPD_DBG(*error_result);
-			fclose(connection->stream);
+			close(connection->socket);
```
sthibaul, line 583: `connection->buf` also needs to be freed

src/api/c/libspeechd.c
```diff
@@ -1788,53 +1806,45 @@ static int spd_set_priority(SPDConnection * connection, SPDPriority priority)
...
 			    ("Error: Can't read reply, broken socket in get_reply!");
-			if (connection->stream != NULL)
-			{
-				fclose(connection->stream);
-				connection->stream = NULL;
-				connection->socket = -1;
-			}
+			if (connection->socket >= 0)
+				close(connection->socket);
+			connection->socket = -1;
```
sthibaul, line 1838: you can keep `connection->socket = -1` in the if

sthibaul:
> This might also fix https://github.com/brailcom/speechd/issues/750

Yes, that will fix it, please also mark it as fixed in the commit log

src/api/c/libspeechd.c
```diff
@@ -517,6 +529,7 @@ SPDConnection *spd_open2(const char *client_name, const char *connection_name,
 				assert(0);
 			SPD_DBG(*error_result);
 			close(connection->socket);
+			free(connection->buf);
```
sthibaul, line 532: ? No, buf is not allocated at this point, and the pointer value is even undefined

sthibaul:
> > This might also fix #750
> 
> Yes, that will fix it, please also mark it as fixed in the commit log

You need to keep "Fixes https://github.com/brailcom/speechd/issues/833" as well

sthibaul:
Thanks a lot!

PR #863, opened 2023-10-27
Add Open JTalk module

tinaxd:
Hello,

This pull request introduces a new module named "openjtalk", using [Open JTalk](https://open-jtalk.sourceforge.net/) for high-quality Japanese text-to-speech synthesis.

To utilize this module, the `open_jtalk` executable, a dictionary, and voice data are required. On Ubuntu, these dependencies can be installed with the following command:
```bash
apt install open-jtalk open-jtalk-mecab-naist-jdic hts-voice-nitech-jp-atr503-m001
```

### DIFF
--- a/config/modules/openjtalk.conf
+++ b/config/modules/openjtalk.conf
@@ -4,7 +4,7 @@ OpenjtalkDictionaryDirectory "/var/lib/mecab/dic/open-jtalk/naist-jdic"
 # HTS voice file (open_jtalk -m option)
 OpenjtalkVoice "/usr/share/hts-voice/nitech-jp-atr503-m001/nitech_jp_atr503_m001.htsvoice"
 
-# Copyright (C) 2023 tinaxd
+# Copyright (C) 2023 Chinamu Kawano <tinaxd@protonmail.com>
 #
 # This program is free software; you can redistribute it and/or modify it under
 # the terms of the GNU General Public License as published by the Free Software
--- a/config/modules/openjtalk.conf
+++ b/config/modules/openjtalk.conf
@@ -1,0 +1,20 @@
+# dictionary directory (open_jtalk -x option)
+OpenjtalkDictionaryDirectory "/var/lib/mecab/dic/open-jtalk/naist-jdic"
+
+# HTS voice file (open_jtalk -m option)
+OpenjtalkVoice "/usr/share/hts-voice/nitech-jp-atr503-m001/nitech_jp_atr503_m001.htsvoice"
+
+# Copyright (C) 2023 tinaxd
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the Free Software
+# Foundation; either version 2 of the License, or (at your option) any later
+# version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+# PARTICULAR PURPOSE.  See the GNU General Public License for more details (file
+# COPYING in the root directory).
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
--- a/configure.ac
+++ b/configure.ac
@@ -376,6 +376,14 @@ AM_CONDITIONAL([kali_support], [test $with_kali != no])
 AM_CONDITIONAL([kali_shim], [test $with_kali = shim])
 AS_IF([test $with_kali != no], [output_modules="${output_modules} kali"])
 
+# check for Open JTalk support
+AC_ARG_WITH([openjtalk],
+	[AS_HELP_STRING([--with-openjtalk], [include Open JTalk support])],
+	[with_openjtalk=yes],
+	[])
+AM_CONDITIONAL([openjtalk_support], [test "$with_openjtalk" = yes])
+AS_IF([test $with_openjtalk = yes], [output_modules="${output_modules} openjtalk"])
+
 audio_dlopen='-dlopen force'
 audio_dlopen_modules=
 
--- a/configure.ac
+++ b/configure.ac
@@ -376,13 +376,9 @@ AM_CONDITIONAL([kali_support], [test $with_kali != no])
 AM_CONDITIONAL([kali_shim], [test $with_kali = shim])
 AS_IF([test $with_kali != no], [output_modules="${output_modules} kali"])
 
-# check for Open JTalk support
-AC_ARG_WITH([openjtalk],
-	[AS_HELP_STRING([--with-openjtalk], [include Open JTalk support])],
-	[with_openjtalk=yes],
-	[])
-AM_CONDITIONAL([openjtalk_support], [test "$with_openjtalk" = yes])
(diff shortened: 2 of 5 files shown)


configure.ac
```diff
@@ -376,6 +376,14 @@ AM_CONDITIONAL([kali_support], [test $with_kali != no])
 AM_CONDITIONAL([kali_shim], [test $with_kali = shim])
 AS_IF([test $with_kali != no], [output_modules="${output_modules} kali"])
 
+# check for Open JTalk support
+AC_ARG_WITH([openjtalk],
+	[AS_HELP_STRING([--with-openjtalk], [include Open JTalk support])],
+	[with_openjtalk=yes],
+	[])
+AM_CONDITIONAL([openjtalk_support], [test "$with_openjtalk" = yes])
+AS_IF([test $with_openjtalk = yes], [output_modules="${output_modules} openjtalk"])
```
sthibaul, line 385: Since this only depends on the openjtalk binaries at run-time, better just always enable the build.

config/modules/openjtalk.conf
```diff
@@ -0,0 +1,20 @@
+# dictionary directory (open_jtalk -x option)
+OpenjtalkDictionaryDirectory "/var/lib/mecab/dic/open-jtalk/naist-jdic"
+
+# HTS voice file (open_jtalk -m option)
+OpenjtalkVoice "/usr/share/hts-voice/nitech-jp-atr503-m001/nitech_jp_atr503_m001.htsvoice"
```
sthibaul, line 5: Will there be other voices? It would be useful to add the series of known possibilities

src/modules/openjtalk.c
```diff
@@ -0,0 +1,237 @@
...
+	DBG("OpenjtalkVoice: %s", OpenjtalkVoice);
+
+	module_audio_set_server();
+
+	return 0;
+}
+
+SPDVoice **module_list_voices(void)
+{
+	/* TODO: Allow users to choose htsvoice */
```
sthibaul, line 78: Yes :)

We can leave as it is for now, but that'll really be useful

src/modules/openjtalk.c
```diff
@@ -0,0 +1,237 @@
...
+
+	DBG("speaking '%s'", data);
+
+	module_report_event_begin();
+
+	/* Strip SSML (Open JTalk does not support it.) */
+	char *plain_data = module_strip_ssml(data);
+
+	char template[] = "/tmp/speechd-openjtalk-XXXXXX";
+	int tmp_fd = mkstemp(template);
```
sthibaul, line 103: better use umask(0077),  to make sure not to leak to other users

src/modules/openjtalk.c
```diff
@@ -0,0 +1,237 @@
...
+		DBG("temporary .wav file creation failed");
+		goto FINISH;
+	}
+
+	/* do not need fd */
+	close(tmp_fd);
+
+	/* construct command line */
+	char cmd[PATH_MAX * 4];
+	snprintf(cmd, sizeof(cmd),
```
sthibaul, line 114: please use asprintf instead

src/modules/openjtalk.c
```diff
@@ -0,0 +1,237 @@
...
+void module_speak_sync(const char *data, size_t bytes, SPDMessageType msgtype)
+{
+	module_speak_ok();
+
+	DBG("speaking '%s'", data);
+
+	module_report_event_begin();
+
+	/* Strip SSML (Open JTalk does not support it.) */
+	char *plain_data = module_strip_ssml(data);
```
sthibaul, line 100: Make sure to release `plain_data` on all exit paths

src/modules/openjtalk.c
```diff
@@ -0,0 +1,237 @@
...
+	AudioFormat format = SPD_AUDIO_LE;
+
+	FILE *audio_fp = fopen(template, "rb");
+	if (audio_fp == NULL) {
+		DBG("failed to open wav file");
+		goto FINISH;
+	}
+	DBG("opened wav file");
+
+	fseek(audio_fp, 34, SEEK_SET);
```
sthibaul, line 152: Please make these 34, 22, etc. macros defined at the top of the file, so they get meaningful naming, and their definitions are in one place.

src/modules/openjtalk.c
```diff
@@ -0,0 +1,237 @@
...
+	/* construct command line */
+	char cmd[PATH_MAX * 4];
+	snprintf(cmd, sizeof(cmd),
+		 "open_jtalk -x %s -m %s -ow %s", OpenjtalkDictionaryDirectory,
+		 OpenjtalkVoice, template);
+
+	FILE *oj_fp = popen(cmd, "w");
+	if (oj_fp == NULL) {
+		DBG("failed to execute open_jtalk");
+		goto FINISH;
```
sthibaul, line 121: I'd say from here on, goto at least to the unlink call, to make sure we don't leak a file in case the program did create the output but things crashed and  `popen` failed for whatever reason.

src/modules/openjtalk.c
```diff
@@ -0,0 +1,237 @@
...
+	DBG("output to %s", template);
+
+	AudioTrack track = {
+		.bits = 0,
+		.num_channels = 0,
+		.sample_rate = 0,
+		.num_samples = 0,
+		.samples = NULL
+	};
+	AudioFormat format = SPD_AUDIO_LE;
```
sthibaul, line 143: Are we sure of this? (just making sure)

sthibaul:
Thanks for this! Just a few comments that should be easy to address.

src/modules/openjtalk.c
tinaxd: It seems that open_jtalk always produces wav files in little endian regardless of machine endianness (confirmed by reading the source code of Open JTalk, as I don't have a PC with big endian).

tinaxd:
Thank you for the review! I have corrected the points you mentioned. I would appreciate if you review my changes.

config/modules/openjtalk.conf
sthibaul: BTW, does the jtalk tool provide a way to get information from a voice file? It'd be better to be able to just configure a directory and let the module look around inside and find the available voices to properly list them.

src/modules/openjtalk.c
sthibaul: ok!

config/modules/openjtalk.conf
tinaxd: I apologize for the late reply.
To the best of my knowledge, open_jtalk doesn't provide a feature to get information from voice files.
Therefore, the most feasible approach would likely be to display the filenames of the .htsvoice files in the specified directory as voice names.

sthibaul: Ok, then a way is to list the available possibilities in the config file with a Dependency line, similarly to the `VoiceFileDependency` option used by the mbrola generic modules. That way the module can check which voices are actually available, while still allowing the user to add more voices to the list.

tinaxd: OK, I will check the implementation of the mbrola module.

sthibaul: the mbrola generic module, more precisely, i.e. the generic module with the `espeak-mbrola-generic.conf` configuration file

PR #865, opened 2023-11-01
Update CLDR to version 44 and other symbols from NVDA and Orca

Nardol:
CLDR version 44.0 has been released and can now be downloaded from unicode.org.
It brings some new emojis and change description for some.
Symbols from actual NVDA beta branch and Orca master has also been imported.


sthibaul:
Thanks!

PR #888, opened 2024-03-18
generic: Abort when speech failed

sthibaul:
Otherwise if the speech command fails, the user just hopelessly gets no speech any more, without speech-dispatcher falling back on something that works.

### DIFF
--- a/src/modules/generic.c
+++ b/src/modules/generic.c
@@ -518,7 +518,7 @@ void *_generic_speak(void *nothing)
 
 				g_free(e_string);
 
-				/* execute_synth_str1 se sem musi nejak dostat */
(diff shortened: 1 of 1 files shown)


Soup-64:
This hardcodes /bin/bash, which can break on systems that either lack bash, or have nonstandard paths like NixOS, which causes the fallback voice to be always used. Why not just do something like `execl("sh", "sh", "-c", command, (char *) NULL)`?

sthibaul:
I had assumed that dash didn't support `set -o pipefail`, but apparently it does, thus changed so in https://github.com/brailcom/speechd/pull/1034

Soup-64:
Awesome, thanks :)

PR #947, opened 2024-07-25
update cldr to 45.0 and unicode

jeremyp3:
I couldn't import the orca symbols because I got a python error: ModuleNotFoundError: No module named 'orca.chnames'.


sthibaul:
Orca does not include symbols any more indeed, it now relies on speech-dispatcher which contains the last orca import that was done from the last version of orca containing symbols :)

PR #948, opened 2024-07-31
Ignore client SSML mode for CHAR and KEY commands

cwendling:
SSML doesn't make sense for those commands, and it can break processing of them.

For example, if a symbol expansion leads to converting it to a TEXT command, it then has to be valid SSML (including surrounding <speak> tags), otherwise a module might not accept it.  If the client is not in SSML mode it worked fine as the input was translated to SSML, but if the client was in that mode the input was still assumed to already be SSML, although it actually wasn't as it's not possible (nor makes sense) to send SSML through CHAR and KEY.

This e.g. broke the Baratinoo module for such symbol expansions (e.g. when "y" is getting translated to "igrèque" to workaround pronunciation issues in some synthesizers).

### DIFF
--- a/src/server/server.c
+++ b/src/server/server.c
@@ -121,6 +121,12 @@ queue_message(TSpeechDMessage * new, int fd, int history_flag,
 		new->time = time(NULL);
 
 		new->settings.paused_while_speaking = 0;
+
+		/* Ignore possible SSML mode for anything but TEXT, as it doesn't make
+		 * sense anything else and can cause problems if the message processing
+		 * assumes SSML although it isn't */
+		if (type != SPD_MSGTYPE_TEXT)
+			new->settings.ssml_mode = SPD_DATA_TEXT;
 	}
 	id = new->id;
 


cwendling:
Hum, just thinking: does SSML mode make sense for *anything* but `TEXT`?  E.g. there's also `SOUND_ICON` and `SPELL`, does it make sense for those?

sthibaul:
It doesn't make sense for them either, it's really just about text

cwendling:
@sthibaul updated accordingly.

src/server/server.c
```diff
@@ -121,6 +121,12 @@ queue_message(TSpeechDMessage * new, int fd, int history_flag,
 		new->time = time(NULL);
 
 		new->settings.paused_while_speaking = 0;
+
+		/* Ignore possible SSML mode for anything but TEXT, as it doesn't make
+		 * sense anything else and can cause problems if the message processing
+		 * assumes SSML although it isn't */
+		if (type != SPD_MSGTYPE_TEXT)
+			new->settings.ssml_mode = SPD_DATA_TEXT;
```
sthibaul, line 129: Why inside the `if (fd>0)` ? a per-uid ssml parameterization would seem odd, but better be safe than sorry?

cwendling: Just because the comment above the function says that *[…] the message _new_ already contains a fully filled in settings structure which should not be overwritten […]*.

I assumed that the message then might already have been partly processed, or that the settings member could be shared between messages so I couldn't restore the SSML flag for TEXT messages needing it.  But all this was merely an assumption seeing all settings-related stuff that was not directly a argument to the function was done there, I didn't actually dig deep enough to understand in which cases it's called with an uid and the detailed implications.
I can change this if needed.

sthibaul: Ok, I see, the caller with nul fd indeed is just re-queueing the message, so it was already processed and fixed, thus no need to fix it again.

Thanks!

PR #950, opened 2024-08-02
Sort Baratinoo engine higher

cwendling:
It was missing from the module order list, leading to have it at the very bottom of the list.  Sort it higher so it can be selected by default in more situations.

Sorting position is somewhat arbitrary, I merely listened to samples for Pico, RHVoice and IVONA and deemed Baratinoo better.  I didn't check all other voices because I didn't easily find relevant samples to listen to, but it's unlikely a user has several of these engines that usually either need specific setup or require buying the corresponding non-free software.


sthibaul:
thanks!

PR #953, opened 2024-08-25
Include note about GenericDelimiters being broken and workaround

guest271314:
Include note about GenericDelimiters being broken and removing silence between sentences.


sthibaul:
It's the texi file which is the source, not the html file.

sthibaul:
Fixed and merged, thanks!

Issue #886, opened 2024-03-17
Have audio backends report playback progress


sthibaul:
**Is your feature request related to a problem? Please describe.**

If the system is somehow loaded, we could get underruns. That's because audio is send chunk by chunk (separated by the indexing marks), so as to follow playback progress.

**Describe the solution you'd like**

The audio modules should instead be able to be fed with all available chunks (that can fit the audio stack buffers), and when playback progress reaches an indexing mark, it would report it.

PR #958, opened 2024-09-23
add initial pipewire audio backend support

albertotirla:
this work is just an initial prototype, this means that some functionality one would normally expect could be missing.

All the required functions are implemented, following the contract set by the other audio backends I had a look at

In particular, setting the volume may or may not work properly, because I'm not sure whether I wrote that part correctly. Furthermore, attempting to set the log level will silently do nothing, because pipewire's systemd service controls the logging. Technically I could switch on the log level for every kind of message I want to log, and then make my own logger implementation for this, but I'm not sure whether it's relevant enough. If it is, I would revisit it. Furthermore, playback may be a bit glitchy, if that happens, do comment here

That being said, this is technically functional, so it should work in practice too. The only reason I'm saying technically here, is because I was unable to properly test it, as the risk of losing speech entirely is too great, and for someone who relies on speech to function on the computer, losing speech is equivalent to a persistent black screen, which is obviously pretty bad.

I wanted to make this pull request now, because the proof of concept is technically ready for testing, and the more I waited, the more my fork would have been behind upstream speech dispatcher

Obviously, code reviews are welcome, especially because I never did anything like this before. So yeah, I may have inadvertently caused a memory leak or something nasty of that sort, so please try to be as rigorous as possible when analyzing this patch. All the linters I could run on this were ran and no issue was found, however that doesn't say much in the realm of C.


sthibaul:
Thanks for contributing this!

From a quick look, it looks good overall, there will be some reworking needed to fix a few issues, but nothing really fundamental to get it working properly enough.

configure.ac
```diff
@@ -29,12 +29,12 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-AC_PREREQ([2.63])
+AC_PREREQ([2.72])
```
sthibaul, line 32: I don't think you really need to bump this, so better not :)

configure.ac
```diff
@@ -29,12 +29,12 @@
...
 m4_define([MAJOR_VERSION], [0])
 m4_define([MINOR_VERSION], [12])
 m4_define([MICRO_VERSION], [0])
 m4_define([AC_PACKAGE_VERSION],[MAJOR_VERSION.MINOR_VERSION.MICRO_VERSION-rc3])
-AC_INIT([speech-dispatcher], [AC_PACKAGE_VERSION], [speechd-discuss@nongnu.org])
+AC_INIT([speech-dispatcher],[AC_PACKAGE_VERSION],[speechd-discuss@nongnu.org])
```
sthibaul, line 37: no need for removing spaces here :)

src/audio/.gitignore
```diff
@@ -23,7 +23,6 @@
 /so_locations
 /spd_alsa.la
 /spd_libao.la
-/spd_nas.la
```
sthibaul, line 26: This is also a spurious removal, it'd rather be adding `/spd_pipewire.la` that would make sense

src/audio/pipewire.c
```diff
@@ -0,0 +1,254 @@
+#ifdef HAVE_CONFIG_H
```
sthibaul, line 1: You need to add the copyright header at the top

src/audio/pipewire.c
```diff
@@ -0,0 +1,254 @@
...
+//  this buffer will be used as backing storage by a ring buffer, which will also insure that the threads are syncronised when reading from and writing to the storage area
+#define SAMPLE_BUFFER_SIZE 16 * 1024 // taken from the pipewire ringbuffer example
+
+// speech dispatcher backend entry point, defined in multiple configurations
+#ifdef USE_DLOPEN
+#define SPD_AUDIO_PLUGIN_ENTRY spd_audio_plugin_get
+#else
+#define SPD_AUDIO_PLUGIN_ENTRY spd_pipewire_LTX_spd_audio_plugin_get
+#endif
+
```
sthibaul, line 28: Please also add the `MSG`/`ERR` macros and put them along the code (and fill set_loglevel), to make bug debunking way easier.

src/audio/pipewire.c
```diff
@@ -0,0 +1,254 @@
...
+    }
+    else
+    {
+        pw_log_error("invalid audio format specifier");
+        return -1;
+    }
+
+    params[0] = spa_format_audio_raw_build(&b, SPA_PARAM_EnumFormat, &SPA_AUDIO_INFO_RAW_INIT(.format = format, .channels = track.num_channels, .rate = track.sample_rate));
+    pw_stream_connect(state->stream, PW_DIRECTION_OUTPUT, PW_ID_ANY, PW_STREAM_FLAG_AUTOCONNECT | PW_STREAM_FLAG_MAP_BUFFERS | PW_STREAM_FLAG_RT_PROCESS, params, 1);
+    pw_thread_loop_start(state->loop);
```
sthibaul, line 160: We don't want to start the thread loop on each begin, but only on `pipewire_open`.

src/audio/pipewire.c
```diff
@@ -0,0 +1,254 @@
...
+        format = SPA_AUDIO_FORMAT_S8;
+    }
+    else
+    {
+        pw_log_error("invalid audio format specifier");
+        return -1;
+    }
+
+    params[0] = spa_format_audio_raw_build(&b, SPA_PARAM_EnumFormat, &SPA_AUDIO_INFO_RAW_INIT(.format = format, .channels = track.num_channels, .rate = track.sample_rate));
+    pw_stream_connect(state->stream, PW_DIRECTION_OUTPUT, PW_ID_ANY, PW_STREAM_FLAG_AUTOCONNECT | PW_STREAM_FLAG_MAP_BUFFERS | PW_STREAM_FLAG_RT_PROCESS, params, 1);
```
sthibaul, line 159: We don't want to connect on each `begin` call, only on first call and when the parameters change (e.g. format difference between several voices)

src/audio/pipewire.c
```diff
@@ -0,0 +1,254 @@
...
+    // to ensure that the params array belongs to us and only we can free it, we will duplicate the string contained in params[1], even if such an operation would have not been necesary in the end
+    state->sink_name = strdup(pars[1]);
+    state->stream = pw_stream_new_simple(pw_thread_loop_get_loop(state->loop), state->sink_name, pw_properties_new(PW_KEY_MEDIA_TYPE, "Audio", PW_KEY_MEDIA_CATEGORY, "Playback", PW_KEY_MEDIA_ROLE, "Accessibility", NULL), &stream_events, &state);
+    return (AudioID *)state;
+}
+// configure a pipewire stream for the given speech dispatcher configuration and then prepair loop for playback
+static int pipewire_begin(AudioID *id, AudioTrack track)
+{
+    module_state *state = (module_state *)id;
+    const struct spa_pod *params[1];
```
sthibaul, line 128: You don't need to make it an array, just `struct spa_po params;` will be fine, and you can pass `&params` to `pw_stream_connect`.

src/audio/pipewire.c
```diff
@@ -0,0 +1,254 @@
...
+    uint8_t buffer[1024];
```
sthibaul, line 129: Is there no way to make sure which size this should be?

src/audio/pipewire.c
```diff
@@ -0,0 +1,254 @@
...
+    }
+    fill_quantity = spa_ringbuffer_get_write_index(&state->rb, &write_index);
+    // we make sure we don't xrun here at least, aka write somehow beyond the beginning of the buffer, or past its end. This is illogical, but we should still ensure that's not the case
+    spa_assert(fill_quantity >= 0);
+    spa_assert(fill_quantity <= SAMPLE_BUFFER_SIZE);
+    // we write the samples provided in the track we were given, assuming num_samples is the actual length in samples of that memory area, to the ringbuffer, to then be consumed by the on_process callback
+    //  we assume speech dispatcher gives us the correct number of bytes for the format it chose, enough for this chunk. If that's not the case, there's not much we could do besides reading uninitialized memory or an incomplete chunk, unfortunately
+    spa_ringbuffer_write_data(&state->rb, state->sample_buffer, SAMPLE_BUFFER_SIZE, write_index, track.samples, track.num_samples);
+    // now, we update the write index to account for the chunk of samples we just wrote
+    spa_ringbuffer_write_update(&state->rb, write_index + track.num_samples);
```
sthibaul, line 181: You also need to add a loop with a condition variable: decrement track.num_samples by the amount that could be filled there, and as long as it is not zero, wait for `on_process` to notify. Otherwise we'll be pushing to the frame_buffer only the first part of the audio that can fit, and not wait for pipewire to  play it before pushing the rest.

src/audio/pipewire.c
```diff
@@ -0,0 +1,254 @@
...
+    }
+    if (must_load_with_silence > 0)
+    {
+        memset(destination_memory + available_for_loading, 0, must_load_with_silence);
+    }
+    // now, we should have most of the data ready, we fill in some metadata and push the buffer object to pipewire
+    buf->datas[0].chunk->offset = 0;
+    buf->datas[0].chunk->size = number_of_frames;
+    buf->datas[0].chunk->stride = stride;
+    pw_stream_queue_buffer(state->stream, b);
```
sthibaul, line 99: You also need to notify pipewire_play, through a condition variable, to tell it that you have freed room in the sample_buffer, so it can fill it again.

src/audio/pipewire.c
```diff
@@ -0,0 +1,254 @@
...
+    pw_thread_loop_destroy(state->loop);
+    // unlink and disconnect the stream
+    pw_stream_disconnect(state->stream);
+    pw_stream_destroy(state->stream);
+    // free the memory allocated by both the duplication of that string in pipewire_begin and the sample buffer allocated to hold samples between pipewire and speech dispatcher
+    free(state->sample_buffer);
+    free(state->sink_name);
+    // uninitialize pipewire
+    pw_deinit();
+    // make audio_id point to null, I dk if there's a way to free that one
```
sthibaul, line 218: `free(state)` does it below

src/audio/pipewire.c
```diff
@@ -0,0 +1,254 @@
...
+    // in the process callback, we take the samples from the ring buffer we used in the pipewire_play function and send it to pipewire for playback
+    // this will be done using the following two step algorythm:
+    // we first get the read index of the ringbuffer, which also returns the amount available in there
+    // if we have enough data in the ringbuffer, aka if the read index isn't less than requested, then we load all of it at once in pipewire
+    // however, if the remainder between the filled part of the buffer and the value of requested is greatter than 0, then we fill that part of the buffer with silence, while taking everything else, hoping we will have more samples next time
+    available_samples = spa_ringbuffer_get_read_index(&state->rb, &read_index);
+    // stride is essentially how many samples there are per channel, in relation to how many channels there are in the output format
+    stride = sizeof(int16_t) * state->channel_count;   // not sure about this one, this could be wrong if the format is 8bit only. This is impractical nowadays, but speech dispatcher still has that case in some modules, so this is left as a mystery and a todo for later
```
sthibaul, line 69: you can easily record in `module_state` the stride (number of bytes per frame), instead of `channel_count`, which you don't really need by itself anyway.

sthibaul (changes requested):
Just a first round of feedback for now

albertotirla (commented):
alright, I made a few changes which addresses some of this, with the rest I'm not understanding clearly how to proceed, therefore they got a question instead of a solved conversation button press

src/audio/pipewire.c
albertotirla: alright, that's now fixed with the latest commit, closing this thread. If there are still issues with the code, feel free to open it again

src/audio/pipewire.c
sthibaul: I'm not asking to use pipewire functions to do the logging: just copy/paste the macro from another module, they will do the right thing.

src/audio/pipewire.c
sthibaul: what do you mean precisely by "multiple times per process": what process are you talking about?

`open` will be called once per output module, either in the module process itself (if the module doesn't support audio through server), or in the spd process (if the module does support audio through server). So there can be several open in the spd process. But that is completely *fine*: the `module_state *` pointer is meant for you to put all you need in it. In practice you don't need to care that open is called several times.

`begin` is called several times, for each synthesis, to tell which parameters are being used (rate, bits, etc.), but not concurrently.

Really, the scenario is *simple*, and you *don't* need to care more than this:

- `open` is called once
- `begin` is called, `feed_sync_overlap` is called several times to provide the pieces of audio as they get synthesized., `end` is called
- another `begin` +  several `feed_sync_overlap` + `end`
- etc.
- `close`, eventually

And the *only* concurrent call that you might encounter is `stop`.

> unless a new instance to the pipewire core object is created as far as I know

Yes, sure, you'll have one core object per `module_state *`.

> I should store them as an array of streams

No you don't, you just store one in your `module_state *`, no more. If you notice in `begin` that the rate/etc. is different from the previous one, just disconnect from the pw stream, and reconnect with updated rate/etc. See how it's done in e.g. the pulse plugin.

> the different copies of the library are mapped in different parts of memory, so they can't touch each others module_state structs

They *don't* need to be in different process to avoid touching to each other's `module_state`. You are *passed* the `module_state` you are supposed to be using.

src/audio/pipewire.c
sthibaul: Ah, I missed that `spa_format_audio_raw_build` returns an `spa_pod*`, not `spa_pod`.

So you can use

```
struct pa_pod *params;
[...]
pw_stream_connect(..., &params, 1);
```

src/audio/pipewire.c
sthibaul: > spd generates enough samples at once to fill that entire buffer?

Of course: speech synthesis produce audio much faster than the audio card plays it.

> how many samples would I typically get at once?

Depends on the syntheses. That's why you shouldn't try to guess and just push to your `frame_buffer` whatever can fit, and then wait for more room to free up before pushing more, etc. until it's all pushed.

> what do you mean by decrementing track.num_samples, I can actually modify that and spd sees it?

No, I rather meant you can make your own copy of it and decrement it.

> I thought that struct is supposed to pass to me as read-only

It's not marked as const, so you can actually modify it. But possibly const was just forgotten, so better make your own copy of the field.

> What should I decrement it by?

The amount you were able to push into `frame_buffer`, i.e. the minimum between its value and  fill_quantity.

src/audio/pipewire.c
sthibaul: > as long as pipewire_play gets called every time there's a new chunk to be played, it will do the right thing, because it automatically fills the buffer with the data

No, it needs to block when the `frame_buffer` is full. And then needs to be woken up once it got some room.

> This would lead to data races if the ringbuffer implementation wasn't guarding access, but it does, so that's not an issue

It does protect access, yes. But it does not provide a blocking primitive. pipewire_play does need to block at some point, otherwise it doesn't know how fast it's supposed to push data to the `frame_buffer`.

src/audio/.gitignore
```diff
@@ -23,8 +23,9 @@
...
 /spd_oss.la
 /spd_pulse.la
+/spd_nas.la
```
sthibaul, line 28: keep it in alphabetical order please

sthibaul:
Also, please add to `.github/workflows/ci.yml` the `libpipewire-0.3-dev` build-dep along `libpulse-dev` (in two places), so that your got gets compiled by the CI

src/audio/pipewire.c
albertotirla: alright, finally, now I understand how this is supposed to work! figuring out how to solve those segmentation faults I'm seeing, then will be doing that

src/audio/.gitignore
albertotirla: ok, done!

src/audio/pipewire.c
```diff
@@ -0,0 +1,277 @@
...
+    if ((destination_memory = buf->datas[0].data) == NULL)
+        return;
+    // in the process callback, we take the samples from the ring buffer we used in the pipewire_play function and send it to pipewire for playback
+    // this will be done using the following two step algorythm:
+    // we first get the read index of the ringbuffer, which also returns the amount available in there
+    // if we have enough data in the ringbuffer, aka if the read index isn't less than requested, then we load all of it at once in pipewire
+    // however, if the remainder between the filled part of the buffer and the value of requested is greatter than 0, then we fill that part of the buffer with silence, while taking everything else, hoping we will have more samples next time
+    available_samples = spa_ringbuffer_get_read_index(&state->rb, &read_index);
+    number_of_frames = buf->datas[0].maxsize / state->stride; // how many frames of audio are in this chunk, aka how many channel_count touples are available this call to on_process. Technically, requested itself could be used, but this is done instead to avoid some very rare, but important, edge cases
```
sthibaul, line 87: If there is an edge case that is important, it should be explained in the comments :)

sthibaul:
There is one issue: mixing number of bytes and number of samples. On malloc, `SAMPLE_BUFFER_SIZE` is used as number of bytes, while on ringbuffer operations it is used as number of frames. That needs to be fixed, otherwise buffer overflows will happen. It needs to be clear everywhere whether we are talking about a number of frames or about  a number of bytes. I would tend to says that it'll be simpler to just always use a number of bytes, and multiple the number of frames very early when getting audio from spd

sthibaul:
Only now I realize: do not implement `play` but `feed_sync_overlap`, so that we have some audio overlap to avoid underruns.

sthibaul:
> Only now I realize: do not implement `play` but `feed_sync_overlap`, so that we have some audio overlap to avoid underruns.

I mean: you can keep your `pipewire_play` as it is, just call it `pipewire_feed_sync_overlap` and put it in the `feed_sync_overlap` field of `spd_audio_plugin_t`

src/audio/pipewire.c
```diff
@@ -0,0 +1,277 @@
...
+    if (b->requested > 0)
+    {
+        number_of_frames = SPA_MIN(number_of_frames, b->requested);
+    }
```
sthibaul, line 91: Apparently `b->requested` was introduced in version 0.3.49 only, not available in CI. So you need to surround this piece with `#if PW_CHECK_VERSION(0, 3, 49)`

src/audio/pipewire.c
```diff
@@ -0,0 +1,277 @@
...
+    must_load_with_silence = number_of_frames - available_for_loading;
+    if (available_for_loading > 0)
+    {
+        spa_ringbuffer_read_data(&state->rb, state->sample_buffer, SAMPLE_BUFFER_SIZE, read_index, destination_memory, available_for_loading);
+        spa_ringbuffer_read_update(&state->rb, read_index + available_for_loading);
+    }
+    if (must_load_with_silence > 0)
+    {
+        memset(destination_memory + available_for_loading, 0, must_load_with_silence);
+    }
```
sthibaul, line 111: I would also say to surround this with `#if PW_CHECK_VERSION(0, 3, 49)` and `if (b->requested)` : it's only when we cannot meet the `requested` constraint that we have to introduce silence. If we don't have such a constraint or cannot know it, we'd fall back to completely filling the buffer, but that might be way more than we actually need, and spuriously empty the `sample_buffer`, leading to introduce silence while we had ample time to buffer that.

src/audio/pipewire.c
sthibaul: yes, you just need to record what the params were last time, no biggie :)

albertotirla (commented):
alright, fixed the easier issues now, to make room for the harder ones

src/audio/pipewire.c
sthibaul: we also want to test `if (b->requested)`. If `requested` is 0, we'll try to fill the whole buffer in, which would be excessive and lead to spurious silence

src/audio/pipewire.c
sthibaul: Like you did it is alright

src/audio/pipewire.c
albertotirla: fixed in the latest commit

src/audio/pipewire.c
albertotirla: fixed in latest commit

src/audio/pipewire.c
albertotirla: fixed in latest commit, hopefully

albertotirla:
alright, this fixes the following things:

* now we're using play instead of feed_sink_overlap
* conditional compilation for older pipewire versions
* throttling so that we don't generate more audio than pipewire can play, overflowing the buffer

But it still segfaults for now

albertotirla:
> There is one issue: mixing number of bytes and number of samples. On malloc, `SAMPLE_BUFFER_SIZE` is used as number of bytes, while on ringbuffer operations it is used as number of frames. That needs to be fixed, otherwise buffer overflows will happen. It needs to be clear everywhere whether we are talking about a number of frames or about a number of bytes. I would tend to says that it'll be simpler to just always use a number of bytes, and multiple the number of frames very early when getting audio from spd

that's now fixed

src/audio/pipewire.c
```diff
@@ -0,0 +1,322 @@
...
+// pipewire on process callback
+// this function would be called each time the server requests data from us
+static void on_process(void *userdata)
+{
+    module_state *state = (module_state *)userdata;
+    // a lot of the code below has been taken from the playing a tone pipewire tutorial, see the pipewire documentation for more details
+
+    struct pw_buffer *b;
+    struct spa_buffer *buf;
+    uint32_t *destination_memory, read_index, available_for_loading, must_load_with_silence;
```
sthibaul, line 74: destination_memory should be a void*, like `buf->datas[0].data` is, not a uint32_t *. Otherwise the computation in  `memset(destination_memory + available_for_loading, 0, must_load_with_silence);` will be completely wrong

src/audio/pipewire.c
```diff
@@ -0,0 +1,322 @@
...
+    {
+        if (must_load_with_silence > 0)
+        {
+            memset(destination_memory + available_for_loading, 0, must_load_with_silence);
+        }
+    }
+#endif
+    // now, we should have most of the data ready, we fill in some metadata and push the buffer object to pipewire
+    buf->datas[0].chunk->offset = 0;
+    buf->datas[0].chunk->size = number_of_samples;
```
sthibaul, line 128: no, it's `available_for_loading` bytes that we have loaded in `buf`.

src/audio/pipewire.c
```diff
@@ -0,0 +1,322 @@
...
+    // if we have enough data in the ringbuffer, aka if the read index isn't less than requested, then we load all of it at once in pipewire
+    // however, if the remainder between the filled part of the buffer and the value of requested is greatter than 0, then we fill that part of the buffer with silence, while taking everything else, hoping we will have more samples next time
+    available_samples = spa_ringbuffer_get_read_index(&state->rb, &read_index);
+    number_of_samples = buf->datas[0].maxsize; // how many samples, in bytes,  of audio are in this chunk. Technically, requested itself could be used, but this is done instead to avoid some very rare, but important, edge cases, for example buggy hardware drivers which don't probe the card correctly, so on_process gets called very often, with a value of requested samples greatter than what pipewire could allocate, in maxsize
+#if PW_CHECK_VERSION(0, 3, 49)
+    if (b->requested > 0)
+    {
+        number_of_samples = SPA_MIN(number_of_samples, b->requested * state->stride); // because this value is given in frames, and we want samples
```
sthibaul, line 96: `samples` is a synonym of `frame`. You need to use `bytes` instead of `samples`.

src/audio/pipewire.c
```diff
@@ -0,0 +1,322 @@
...
+{
+    module_state *state = (module_state *)malloc(sizeof(module_state));
+    pw_init(0, NULL);
+    state->loop = pw_thread_loop_new("pipewire audio thread", NULL);
+    state->rb = SPA_RINGBUFFER_INIT();
+    state->sample_buffer = malloc(SAMPLE_BUFFER_SIZE);
+    // we don't know if the params array belongs to us
+    // it  could be freed by speech dispatcher at a later date, which means we could be storing a pointer to freed memory
+    // to ensure that the params array belongs to us and only we can free it, we will duplicate the string contained in params[1], even if such an operation would have not been necesary in the end
+    state->sink_name = strdup(pars[1]);
```
sthibaul, line 153: `pw_stream_new_simple` makes a copy of the `name` being passed, so no need to strdup it.

src/audio/pipewire.c
```diff
@@ -0,0 +1,322 @@
...
+    pw_thread_loop_lock(state->loop);
+    pw_stream_connect(state->stream, PW_DIRECTION_OUTPUT, PW_ID_ANY, PW_STREAM_FLAG_AUTOCONNECT | PW_STREAM_FLAG_MAP_BUFFERS | PW_STREAM_FLAG_RT_PROCESS, &params, 1);
+    pw_thread_loop_unlock(state->loop);
+    return 0;
+}
+static int pipewire_feed_sink_overlap(AudioID *id, AudioTrack track)
+{
+    module_state *state = (module_state *)id;
+    uint32_t write_index, fill_quantity, room_left_in_buffer = 0, track_buffer_size;
+    uint64_t dummy; // used to read from the event file descripter. We don't use this for anything, at the moment, it's always 42
```
sthibaul, line 209: Actually it will be more than 42 if `on_process` is called several times before `pipewire_feed_sink_overlap` gets to execute.

src/audio/pipewire.c
```diff
@@ -0,0 +1,322 @@
...
+    // if the stream has been deactivated because of a pipewire_stop call, we activate it
+    pw_thread_loop_lock(state->loop);
+    if (pw_stream_get_state(state->stream, NULL) == PW_STREAM_STATE_PAUSED)
+    {
+        pw_stream_set_active(state->stream, true);
+    }
+    pw_thread_loop_unlock(state->loop);
+    track_buffer_size = track.num_samples;
```
sthibaul, line 217: you need to multiply by the stride to get a number of bytes

src/audio/pipewire.c
```diff
@@ -0,0 +1,322 @@
...
+    {
+        // drainage spinlock!
+        // we wait till fill_quantity is not empty after this thread can continue, aka after on_process updated that ringbuffer index with what it managed to play, so that we have some room in the buffer to fill from what's remaining of what was given to us by speech dispatcher
+        while (true)
+        {
+            fill_quantity = spa_ringbuffer_get_write_index(&state->rb, &write_index);
+            // see if there's some room in the buffer, if so, stop spinning
+            room_left_in_buffer = SAMPLE_BUFFER_SIZE - fill_quantity;
+            if (room_left_in_buffer > 0)
+                // then we have room in the buffer, so stop spinning
```
sthibaul, line 229: Do not call it "spinning": we are *not* spinning but blocking

src/audio/pipewire.c
```diff
@@ -0,0 +1,322 @@
...
+                break;
+            // otherwise, to not starve a cpu core of resources, we suspend our spinning until something, in this case on_process, wakes us up. In an embedded system without these primitives, spinning endlessly like that would be the only option. We could do the same here, but for efficiency reasons alone, we don't
+            pw_thread_loop_lock(state->loop);
+            struct pw_loop *program_loop = pw_thread_loop_get_loop(state->loop);
```
sthibaul, line 233: you could record the `pw_loop` in `state` too, to avoid having to lock each time

src/audio/pipewire.c
```diff
@@ -0,0 +1,322 @@
...
+            pw_thread_loop_unlock(state->loop);
+            spa_system_eventfd_read(program_loop->system, state->eventfd_number, &dummy);
+        }
+        // we write the amount of samples we can at this time without overflowing the buffer, to the memory area represented by the amount of room that's free after the last on_process call, to then be enqueued by pipewire
+        //  we assume speech dispatcher gives us the correct number of bytes for the format it chose, enough for this chunk. If that's not the case, there's not much we could do besides reading uninitialized memory or an incomplete chunk, unfortunately
+        // if we are about to overflow the buffer, better sharply clamp it than cause a buffer overflow vulnerability
+        room_left_in_buffer = SPA_MIN(SAMPLE_BUFFER_SIZE, room_left_in_buffer);
```
sthibaul, line 240: There is no need for this `SPA_MIN` call: by construction `room_left_in_buffer` cannot be greater than `SAMPLE_BUFFER_SIZE`

src/audio/pipewire.c
```diff
@@ -0,0 +1,322 @@
...
+    }
+    else
+    {
+        available_for_loading = 0;
+    }
+    // if there's anything else to fill and the ring buffer doesn't contain enough at this point, fill the difference with silence
+    must_load_with_silence = number_of_samples - available_for_loading;
+    if (available_for_loading > 0)
+    {
+        spa_ringbuffer_read_data(&state->rb, state->sample_buffer, SAMPLE_BUFFER_SIZE, read_index, destination_memory, available_for_loading);
```
sthibaul, line 113: `read_index` needs a `%SAMPLE_BUFFER_SIZE`, otherwise you will be reading garbage

src/audio/pipewire.c
```diff
@@ -0,0 +1,322 @@
...
+            pw_thread_loop_unlock(state->loop);
+            spa_system_eventfd_read(program_loop->system, state->eventfd_number, &dummy);
+        }
+        // we write the amount of samples we can at this time without overflowing the buffer, to the memory area represented by the amount of room that's free after the last on_process call, to then be enqueued by pipewire
+        //  we assume speech dispatcher gives us the correct number of bytes for the format it chose, enough for this chunk. If that's not the case, there's not much we could do besides reading uninitialized memory or an incomplete chunk, unfortunately
+        // if we are about to overflow the buffer, better sharply clamp it than cause a buffer overflow vulnerability
+        room_left_in_buffer = SPA_MIN(SAMPLE_BUFFER_SIZE, room_left_in_buffer);
+        spa_ringbuffer_write_data(&state->rb, state->sample_buffer, SAMPLE_BUFFER_SIZE, write_index % SAMPLE_BUFFER_SIZE, track.samples, room_left_in_buffer);
+        // subtract from the total buffer what we were already able to write, in order to indicate that the buffer  has more room
```
sthibaul, line 242: no, it's not to indicate that the buffer has more room, but rather that we have progressed in pushing the audio to `sample_buffer`

src/audio/pipewire.c
sthibaul: You however want to call `SPA_MIN` with `track_buffer_size`, to handle the case when we have less audio to push than the available room. Otherwise you'll be pushing garbage.

src/audio/pipewire.c
```diff
@@ -0,0 +1,322 @@
...
+        track_buffer_size -= room_left_in_buffer;
+        // now, we update the write index to account for the chunk of samples we just wrote
+        spa_ringbuffer_write_update(&state->rb, write_index + room_left_in_buffer);
+    }
```
sthibaul, line 246: After this loop, you will want a loop that will wait for `spa_ringbuffer_get_write_index` to return a value that is below some threshold, such as 20ms worth of audio, i.e. `sample_rate * stride * 20 / 1000`, to achieve the targetted overlap: not too much so that index reporting is precise, but not too little for spd to have the time to push the subsequent audio.

Ideally, later we'll be able to make audio plugins just report the audio progression, see https://github.com/brailcom/speechd/issues/886

sthibaul (changes requested):

src/audio/pipewire.c
albertotirla: fixed

src/audio/pipewire.c
```diff
@@ -0,0 +1,322 @@
...
+    // pipewire's buffers are mapped directly from the server to the client
+    // As such, if the memory we're given doesn't exist, we have to abort imediately, because doing otherwise most certainly would lead to undefined behaviour
+    if ((destination_memory = buf->datas[0].data) == NULL)
+        return;
+    // in the process callback, we take the samples from the ring buffer we used in the pipewire_play function and send it to pipewire for playback
+    // this will be done using the following two step algorythm:
+    // we first get the read index of the ringbuffer, which also returns the amount available in there
+    // if we have enough data in the ringbuffer, aka if the read index isn't less than requested, then we load all of it at once in pipewire
+    // however, if the remainder between the filled part of the buffer and the value of requested is greatter than 0, then we fill that part of the buffer with silence, while taking everything else, hoping we will have more samples next time
+    available_samples = spa_ringbuffer_get_read_index(&state->rb, &read_index);
```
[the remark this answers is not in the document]
albertotirla: or rather said, should I rename that variable?

src/audio/pipewire.c
albertotirla: fixed

src/audio/pipewire.c
albertotirla: I will close this thread now, if that clarifies another part of the code and there's a bug there or something, feel free to reopen

src/audio/pipewire.c
albertotirla: done

src/audio/pipewire.c
albertotirla: how do you mean? so, after both loops close, something like this?

```C
available_samples=0; //reset
while(available_samples<=track.sample_rate*state->stride*20/1000) //20 ms of audio accumulated in the ringbuffer, but...who sends it if our thread is blocked?
    {
    available_samples = spa_ringbuffer_get_write_index(&state->rb, &write_index); //update the value
    //perhaps wait for the eventfd again? dk how exactly this is supposed to work
    }
```

I'm unsure what you mean, but yeah, like that?

src/audio/.gitignore
```diff
@@ -23,8 +23,9 @@
 /so_locations
 /spd_alsa.la
 /spd_libao.la
-/spd_nas.la
 /spd_oss.la
 /spd_pulse.la
+/spd_pipewire.la
+/spd_nas.la
```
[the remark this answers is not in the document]
albertotirla: ok, fixed this time, hopefully ⭕

src/audio/pipewire.c
```diff
@@ -0,0 +1,322 @@
...
+    if ((b = pw_stream_dequeue_buffer(state->stream)) == NULL)
+    {
+        pw_log_warn("out of buffers: %m");
+        return;
+    }
+    buf = b->buffer;
+    // pipewire's buffers are mapped directly from the server to the client
+    // As such, if the memory we're given doesn't exist, we have to abort imediately, because doing otherwise most certainly would lead to undefined behaviour
+    if ((destination_memory = buf->datas[0].data) == NULL)
+        return;
```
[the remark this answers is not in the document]
sthibaul: Then really abort() here. Otherwise spd will just stay *stuck* instead of restarting, and thus completely lose speech.

src/audio/pipewire.c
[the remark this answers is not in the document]
sthibaul: yes

src/audio/pipewire.c
sthibaul: I meant fix the comment

src/audio/pipewire.c
sthibaul: Ah, the name of the variable `number_of_samples` mislead me, I was thinking it was always the whole buffer. But in the `(!b->requested)` case (or old pipewire version), you do want to use `available_for_loading`, since in that case you won't have filled with silence.

src/audio/pipewire.c
sthibaul: It is sort of a dummy value, yes, but it still adds up each time on_process is called, and will thus be a multiple of 42.

src/audio/pipewire.c
sthibaul: That'd be really meaningless here: you're releasing the lock before using `program_loop`. If `program_loop` was ever to change, you'd be screwed right after you unlock. Pipewire allocates the program loop on `pw_thread_loop_new`, and that never changes after that.

src/audio/pipewire.c
sthibaul: you do want  to wait for eventfd, yes, and the while is the contrary: it's as long as the buffer is filled with audio that we want to wait.

src/audio/pipewire.c
sthibaul: Particularly for embedded systems you do not want to spin and suck battery

sthibaul (changes requested):

src/audio/pipewire.c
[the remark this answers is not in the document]
albertotirla: can I call exit from here? how would I abort, except manually causing a segmentation fault intentionally?

src/audio/pipewire.c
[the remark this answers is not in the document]
albertotirla: fixed in latest commit

src/audio/pipewire.c
albertotirla: I don't much care about that kind of old pipewire, if it's old enough to not have this function, it's very buggy and barely usable anyway. I may add it in the future, but we have to get actual speech out of this

src/audio/pipewire.c
albertotirla: alright, will do then

src/audio/pipewire.c
albertotirla: ahh, so > instead of <, ok

src/audio/pipewire.c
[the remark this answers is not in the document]
sthibaul: Like I said: really `abort()`.

src/audio/pipewire.c
sthibaul: It's not just about old pipewire, but also the case when a new pipewire doesn't provide a `b->requested`

sthibaul: Also, it's not that old: the previous ubuntu LTS, 22.04, has that old version of pipewire

src/audio/pipewire.c
```diff
@@ -0,0 +1,318 @@
...
+        }
+        // we write the amount of samples we can at this time without overflowing the buffer, to the memory area represented by the amount of room that's free after the last on_process call, to then be enqueued by pipewire
+        //  we assume speech dispatcher gives us the correct number of bytes for the format it chose, enough for this chunk. If that's not the case, there's not much we could do besides reading uninitialized memory or an incomplete chunk, unfortunately
+        // in case we produced more audio than the available room in the buffer that got freed by on_process, we only push what we can, otherwise we would read from uninitialised memory
+        room_left_in_buffer = SPA_MIN(track_buffer_size, room_left_in_buffer);
+        // everything has been accounted for, so write the first batch of data, keeping into account that we can't overflow the initially allocated block for the buffer, represented by SAMPLE_BUFFER_SIZE
+        spa_ringbuffer_write_data(&state->rb, state->sample_buffer, SAMPLE_BUFFER_SIZE, write_index % SAMPLE_BUFFER_SIZE, track.samples, room_left_in_buffer);
```
sthibaul, line 238: track.samples is incorrect here, it's always the same on each loop iteration. You need to use a variable that you advance

src/audio/pipewire.c
```diff
@@ -128,8 +128,8 @@ static void on_process(void *userdata)
     buf->datas[0].chunk->size = number_of_bytes;
     buf->datas[0].chunk->stride = state->stride;
     pw_stream_queue_buffer(state->stream, b);
-    // signal to the main thread that data can be written again
-    spa_system_eventfd_write(pw_thread_loop_get_loop(state->loop)->system, state->eventfd_number, 42);
+    // signal to the main thread that data can be written again, make sure it's woken only once
+    spa_system_eventfd_write(pw_thread_loop_get_loop(state->loop)->system, state->eventfd_number, 1);
```
sthibaul: ? This is not a number of wakeups, since you have not passed the SPA_FD_EVENT_SEMAPHORE flag on event creation

src/audio/pipewire.c
albertotirla: this is not needed after all, according to my preliminary testing

src/audio/pipewire.c
albertotirla: then, does the value matter? I put 1 in there now, so I don't get as many client missed 1 wakeups warnings, though I still do from time to time. As far as I remember, I deleted the 42 part of the comment

src/audio/pipewire.c
sthibaul: That's probably only by luck. Try increasing the buffer size (which can be very useful to avoid underruns), and you will get some shift in the index reporting

src/audio/pipewire.c
sthibaul: The value does not matter since you don't use it when waiting for the event.

> I put 1 in there now, so I don't get as many client missed 1 wakeups warnings

Did you really measure a difference, or is that just an impression?

src/audio/pipewire.c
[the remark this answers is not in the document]
albertotirla: done!

src/audio/pipewire.c
albertotirla: that's fixed now, kinda. We have no way of knowing what pw expects of us in that situation, so we make a best guess by pushing what we have, better than pushing nothing

src/audio/pipewire.c
albertotirla: tryed that, didn't make a difference. From 16*1024 to 64*1024

src/audio/pipewire.c
albertotirla: this is finally fixed in the last commit, closing this as done

src/audio/pipewire.c
```diff
@@ -0,0 +1,419 @@
...
+// if our pipewire is new enough, we know exactly how many samples to push, given through the value of the requested field, from the structure returned when we dequeue a buffer for filling
+#if PW_CHECK_VERSION(0, 3, 49)
+    if (b->requested > 0)
+    {
+        total_number_of_bytes_in_pipewire_buffer = SPA_MIN(total_number_of_bytes_in_pipewire_buffer, b->requested * state->stride); // because this value is given in frames/samples, and we want bytes
+    }
+// otherwise however, we have no idea how much we should push, so we push as much as we have in the ring buffer, making sure we don't push more than maxsize. If that ends up being less than what would have been requested if we had access to that, we have no choice but to underrun
+#else
+    total_number_of_bytes_in_pipewire_buffer = SPA_MIN(available_bytes_in_speech_dispatcher_buffer, buf->datas[0].maxsize * state->stride); // bytes, not frames)
```
sthibaul, line 119: You don't need to refer to maxsize again (and it's bogus here, you shouldn't multiply by stride), it's already used to initialize `total_number_of_bytes_in_pipewire_buffer`. Just use `SPA_MIN(available_bytes_in_speech_dispatcher_buffer, total_number_of_bytes_in_pipewire_buffer)`

src/audio/pipewire.c
```diff
@@ -0,0 +1,419 @@
...
+        available_for_loading = SPA_MIN(available_bytes_in_speech_dispatcher_buffer, total_number_of_bytes_in_pipewire_buffer);
+        message(4, "we can load %i bytes for now", available_for_loading);
+    }
+    else
+    {
+        available_for_loading = 0;
+        message(4, "nothing to be loaded");
+    }
+    // if there's anything else to fill and the ring buffer doesn't contain enough at this point, fill the difference with silence
+    must_load_with_silence = total_number_of_bytes_in_pipewire_buffer - available_for_loading;
```
sthibaul, line 134: You can move this computation inside `if (b->requested > 0)`

src/audio/pipewire.c
```diff
@@ -0,0 +1,419 @@
...
+        message(4, "backend initialised with different sample rate, another module connecting?");
+        message(4, "disconnecting the old stream now");
+        pw_stream_disconnect(state->stream);
+        // but don't forget to update the value of sample rate we cache, for next time
+        state->playback_sample_rate = track.sample_rate;
+    }
+    // otherwise, we went here before, but we got an identical sample rate, why are we called by spd? this must be a spurious wakeup
+    else
+    {
+        message(3, "warning: spurious call by speech dispatcher, sample rate is the same as what this backend was initialised with");
```
sthibaul, line 256: It's not spurious, it's only normal: begin is called for each utterance, and end is called when it's over (i.e. spd will possibly not be sending audio for some time). When that happens, do not connect again, you can just re-use the existing connected stream. See how other audio backends do it.

src/audio/pipewire.c
```diff
@@ -0,0 +1,419 @@
...
+                message(4, "we have %i bytes of free room in buffer, stop locking", room_left_in_buffer);
+                break;
+            }
+            // otherwise, to not starve a cpu core of resources, we suspend our spinning until something, in this case on_process, wakes us up
+            message(4, "no space, waiting for wakeup");
+            spa_system_eventfd_read(state->inner_loop->system, state->eventfd_number, &dummy);
+        }
+        // we write the amount of samples we can at this time without overflowing the buffer, to the memory area represented by the amount of room that's free after the last on_process call, to then be enqueued by pipewire
+        //  we assume speech dispatcher gives us the correct number of bytes for the format it chose, enough for this chunk. If that's not the case, there's not much we could do besides reading uninitialized memory or an incomplete chunk, unfortunately
+        // in case we produced more audio than the available room in the buffer that got freed by on_process, we only push what we can, otherwise we would read from uninitialised memory
```
sthibaul, line 314: The comment is not coherent: when there is more audio to push than available room, we wouldn't be reading unitialized memory, but overwriting audio being played

sthibaul: You probably rather meant: when there is less audio to push than available room

config/speechd.conf
```diff
@@ -197,16 +197,22 @@ SymbolsPreprocFile "symbols-fallback.dic"
...
-# include Pulse Audio. NAS provides network transparency, but is not
-# very well tested. libao is a cross platform library with plugins for
+# include Pipe Wire. NAS provides network transparency, but is not
+# very well tested.
+# libao is a cross platform library with plugins for
 # different sound systems and provides alternative output for Pulse Audio
 # and ALSA as well as for other backends.
 
-# AudioOutputMethod "pulse"
+AudioOutputMethod "pipewire"
```
sthibaul, line 215: I don't think we want to make it the default yet. Better have it tested by beta-testers before enabling it for everybody.

src/audio/pipewire.c
albertotirla: yep, that's what I ment indeed, sorry for the confusion

src/audio/pipewire.c
albertotirla: so, the must_load_with_silence variable? or the whole block, where the dependencies of that variable are calculated?

src/audio/pipewire.c
albertotirla: alright, corrected to something which should be more coherent. To clarify, doing nothing and returning is the best thing to do still, even if my comment was wrong?

config/speechd.conf
albertotirla: done, I also added some stuff about how the transition should be done

src/audio/pipewire.c
sthibaul: I'm only talking about the computation of must_load_with_silence.
available_for_loading and total_number_of_bytes_in_pipewire_buffer are used below so their computation need to be left as such.

src/audio/pipewire.c
sthibaul: Yes

config/speechd.conf
```diff
@@ -197,16 +197,23 @@ SymbolsPreprocFile "symbols-fallback.dic"
...
-# include Pulse Audio. NAS provides network transparency, but is not
-# very well tested. libao is a cross platform library with plugins for
+# include Pipe Wire. NAS provides network transparency, but is not
+# very well tested.
+# libao is a cross platform library with plugins for
 # different sound systems and provides alternative output for Pulse Audio
 # and ALSA as well as for other backends.
 
-# AudioOutputMethod "pulse"
+#audioOutputMethod "pipewire"
```
sthibaul, line 216: Beware of the typo, and the removal of space. This file is used as template for spd-conf to produce the user configuration, so we absolutely need `AudioOutputMethod` to appear as such, and for a space to exist between `#` and `AudioOutputMethod`

alex19EP:
my preliminary quick testing is not very successful. I applied this and installed the built package on Arch Linux, after changing the parameter in the config and restarting sd and orca, I did not hear anything. I'll try to get the logs, but it's difficult to do without a working sd

alex19EP:
hmm. disregard my previous message. I just built sd without libpipewire. It’s a little strange that even after this, pw was still listed as a possible option in the configs. everything works now. although not as good as if I used pulse I hear clicks and other artifacts. especially when navigating quickly or when speaking is paused.

src/audio/pipewire.c
albertotirla: will do...actually, done!

config/speechd.conf
albertotirla: this has now been fixed. Thanks, didn't know spd-conf exists

albertotirla:
re, the configs not changing if you compile without pipewire support:

No idea, I wrote those in speechd.conf, is the configuration adaptive @sthibaul?

re, clicking, crackling and other weird artifacts:

Can you paste a speech dispatcher log when that happens? particularly, increasing the log level to 4 or 5, grepping everything out except pipewire? Furthermore, can you paste the logs you get in stdout/stderr if you run the spd executable with the PIPEWIRE_DEBUG environment variable set to 4?

I sometimes get this too, but very rarely and I thought it's my bad hardware with buggy drivers. I suspect us either not having enough buffering in place, or worse, the locking loop blocking for too long, making pipewire cause our client to xrun. Or, hmm, perhaps 0 is not a good silence value for s16le? could the spd pipeline be too slow for a pull based approach like pipewire, while it wasn't with pulse? or well, obviously there's the consideration that my code is just bad, period, let's never forget that :p

albertotirla:
@alex19EP: I just pushed afew commits to try and fix the crackling. Can you pull and test again?

Warning: system might be a bit sluggish for now. I'm thinking about proper values to wait for, unless the wait for audio to fall below 10 ms logic is actually wrong here, in which case I have to revisit it. Maybe the slight lag is only on my side?

alex19EP:
Yes. There seems to be no more clicking now. Well, or so far I haven’t been able to notice them. I also don't feel any deterioration in performance.

configure.ac
```diff
@@ -420,7 +420,23 @@ AM_CONDITIONAL([libao_support], [test $with_libao = "yes"])
...
+# check for pipewire support
+AC_ARG_WITH([pipewire],
+	[AS_HELP_STRING([--with-pipewire], [include pipewire support])],
+	[],
+	[with_pipewire=check])
+AS_IF([test $with_pipewire != "no"],
+	[PKG_CHECK_MODULES([pipewire], [libpipewire-0.3],
+		[with_pipewire=yes
+		AS_IF([test -z "$default_audio_method"],
+			[default_audio_method=pipewire])
```
sthibaul, line 432: We don't want this precisely :)

sthibaul:
> re, the configs not changing if you compile without pipewire support:
> 
> No idea, I wrote those in speechd.conf, is the configuration adaptive @sthibaul?

the configuration template is not adaptative, no

src/audio/pipewire.c
```diff
@@ -113,6 +113,9 @@ static void on_process(void *userdata)
     if (b->requested > 0)
     {
         total_number_of_bytes_in_pipewire_buffer = SPA_MIN(total_number_of_bytes_in_pipewire_buffer, b->requested * state->stride); // because this value is given in frames/samples, and we want bytes
+        // if there's anything else to fill and the ring buffer doesn't contain enough at this point, fill the difference with silence
+        must_load_with_silence = total_number_of_bytes_in_pipewire_buffer - available_for_loading;
```
sthibaul, line 117: No :) I meant to move it in the other `if (b->requested)` block, where it is actually used. It's no use computing it that early, better compute it just before using it.

src/audio/pipewire.c
```diff
@@ -334,6 +333,8 @@ static int pipewire_stop(AudioID *id)
     message(3, "stopping audio playback");
     // unset the running flag
     state->is_running = false;
+    // drain the stream, to make sure there are no more samples playing, to avoid crackling
+    pw_stream_flush(state->stream, false);
```
sthibaul, line 344: I'm thinking this should probably be done after the pw_stream_set_active call below? otherwise the thread might have time to push some samples after the flush, and thus bringing some crackling (though more rarely). This stop code probably deserves proof-reading from a pipewire guy.

src/audio/pipewire.c
```diff
@@ -324,6 +324,13 @@ static int pipewire_feed_sink_overlap(AudioID *id, AudioTrack track)
         samples_buffer_copy += room_left_in_buffer;
         message(4, "writing completed without errors");
     }
+    // because this is feed_sink_overlap, we wait till the ringbuffer falls below 10 ms, to not lose some small segments of speech
+    fill_quantity = 0;                                                     // reset
+    while (fill_quantity >= track.sample_rate * state->stride * 10 / 1000) // untill there's less than 10 ms in the ringbuffer
+    {
+        fill_quantity = spa_ringbuffer_get_write_index(&state->rb, &write_index); // update the value
```
sthibaul, line 331: you want to update the value *after* calling spa_system_eventfd_read, not before.

And as it is now (fill_quantity initialized to 0), the loop will not actually do anything: fill_quantity being 0, it cannot be larger than 10ms of audio data.

configure.ac
albertotirla: this is how pulse and alsa declared themselves. I have no idea how autotools works, so I copied it. What do we want?

src/audio/pipewire.c
albertotirla: ahh ok, will do, seemned weird to me that that's what you want, I get it now

src/audio/pipewire.c
albertotirla: apparently, stop calls flush on its own, so I will remove that soon. The pipewire person said it's ok how I did it, so yeah, dk what could be bringing on the cracling

configure.ac
sthibaul: Yes, because we want pulse and alsa to be the default if available. But for pipewire, not yet.
Just drop
```
		AS_IF([test -z "$default_audio_method"],
			[default_audio_method=pipewire])
```

src/audio/pipewire.c
albertotirla: oops, weird that I still felt some kind of latency from it though, maybe this is software bug placebo, lol

src/audio/pipewire.c
sthibaul: Which stop?

albertotirla: when I set the state of the stream to inactive, that's like a stop operation, which apparently flushes the stream too

src/audio/pipewire.c
albertotirla: done now in the latest commit

src/audio/pipewire.c
albertotirla: fixed

configure.ac
albertotirla: ok, would people be able to build spd with pipewire support then?

albertotirla: also, can't packagers who don't want that just disable it through flags or something? Either way, I did it, but now I dk how to build my own copy with pipewire support, must I pass something via flags or something? I never used autoconf, so have no idea

sthibaul: > also, can't packagers who don't want that just disable it through flags or something?

Sure they can, but we as upstream do not want people to set it as default audio method yet, to avoid getting a flurry of bug reports, and rather only get reports from a few beta testers.

> now I dk how to build my own copy with pipewire support, must I pass something via flags or something?

You don't need anything, all audio methods are built by default, you just need to choose which one you want in speechd.conf.

sthibaul:
I have merged this, huge thanks!

albertotirla:
it says closed with unmerged commits. Did you close it, or merge it?

sthibaul:
I merged it by hand (fixed a couple formatting issues alongside)

shindere:
I just opened the issue #1031 to report voice quality loss when switching
the output method from pulseaudio to pipewire.

Issue #971, opened 2024-11-21
Crash can occur depending on timing of the clearing of speaking_module


bbergquist0930:
https://github.com/brailcom/speechd/blame/615624a626d78503d09f7937dd56417c96587dfa/src/server/output.c#L969C14-L969C29

```
	if (!speaking_module)
		// We were cancelled
		return;
	pthread_mutex_lock(&playback_events_mutex);
	playback_events = g_slist_append(playback_events, entry);
	pthread_mutex_unlock(&playback_events_mutex);
	ret = write(speaking_module->pipe_speak[1], &c, 1);    // <----- speaking_module can be NULL here
	if (ret != 1)
		MSG(1, "Warning: couldn't write to pipe_speak: %d returned, (errno = %d, %s)\n", ret, errno, strerror(errno));
}
```

This has happened in practice and a core dump occurred and I was able to analyze this.  This `speaking_module` variable is global and is set and cleared by the `speak` thread but is also accessed in the `speak_queue_stop_or_pause`.  This global variable is not protected by any locking, so it is possible that the `speak` thread clears the variable and then the above code accesses the variable.  Even the test for NULL is not going to help because a thread swap could occur after that.

This really needs to have a lock.

sthibaul:
Fixed by #972

PR #972, opened 2024-11-22
Issue #971 Make local copy of speeking_module to ensure it can be tested for NULL

bbergquist0930:
The `speak` thread can change the `speeking_module` to NULL while we are in this function.  So even though `speeking_module` is tested for NULL, this is not good enough in a multi-threading environment.

This change grabs a local copy of `speeking_module` and then tests that for NULL so that even if a thread context switch were to occur and `speeking_module` is set to NULL, the local copy will remain stable and usable.

### DIFF
--- a/src/server/output.c
+++ b/src/server/output.c
@@ -960,15 +960,13 @@ static void output_queue_event(speak_queue_entry *entry)
 {
 	char c = 0;
 	int ret;
-	if (!speaking_module)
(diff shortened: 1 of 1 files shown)


sthibaul:
Please merge the test for NULL with the test above, no need to have two of them :)

sthibaul:
thanks!

PR #978, opened 2024-12-18
espeak-ng-mbrola-generic: Fix missing sound at the end of messages

cwendling:
At least espeak-ng 1.51 from Debian and Ubuntu eat the last character from stdin, so we need to pad with a safe dummy for the whole expected input to be processed.

All is good with espeak (non-ng) 1.48.15:

```shell
$ printf %s 'Bonjour' | espeak -v mb-fr1 -s 160 -p 50  -q --stdin --pho
b	70
o~	41	 0 94 20 95 40 96 59 97 80 99 100 99
Z	70
u	74	 0 102 80 76 100 76
R	70
_	350
_	1
```

But not so much with espeak-ng 1.51:

```shell
$ printf %s 'Bonjour' | espeak-ng -v mb-fr1 -s 160 -p 50  -q --stdin --pho
b	70
o~	41	 0 94 20 95 40 96 59 97 80 99 100 99
Z	70
u	120	 0 102 80 76 100 76
_	350
_	1
```

Notice the missing `R`.

Now, if we pad the input with a single space:

```shell
$ printf '%s ' 'Bonjour' | espeak-ng -v mb-fr1 -s 160 -p 50  -q --stdin --pho
b	70
o~	41	 0 94 20 95 40 96 59 97 80 99 100 99
Z	70
u	74	 0 102 80 76 100 76
R	70
_	350
_	1
```

So, do so in espeak-ng-mbrola-generic not to stumble on the issue. Adding one trailing space is likely safe enough that it will not start causing issues with future fixed version, and testing showed it doesn't affect the tested version: adding one or more spaces yielded the exact same results for both espeak and espeak-ng with the tested versions.

### DIFF
--- a/config/modules/espeak-ng-mbrola-generic.conf
+++ b/config/modules/espeak-ng-mbrola-generic.conf
@@ -22,13 +22,13 @@ Debug 0
 # can modify this value, see other parameters).
 # The command can be split into more lines, if necessary, using '\'.
 GenericExecuteSynth \
-"printf %s \'$DATA\' | espeak-ng -v mb-$VOICE -s $RATE -p $PITCH $PUNCT -q --stdin --pho | mbrola -v $VOLUME -e /usr/share/mbrola/$VOICE/$VOICE - -.au | $PLAY_COMMAND"
+"printf \'%s \' \'$DATA\' | espeak-ng -v mb-$VOICE -s $RATE -p $PITCH $PUNCT -q --stdin --pho | mbrola -v $VOLUME -e /usr/share/mbrola/$VOICE/$VOICE - -.au | $PLAY_COMMAND"
 
 # Alternatively you can shorten the command like below, which makes it
 # work directly with any audio playback utility, but then you won't
 # be able to change the volume from the client application:
 # GenericExecuteSynth \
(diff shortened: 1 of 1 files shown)


sthibaul:
Does it not affect single-letter pronunciation? (for e.g. spelling)

sthibaul:
Is the issue reported to espeak-ng?

sthibaul:
I bisected between 1.48.15 and 1.49, the "culprit" was eaa0c9aa08350ce8a1343aed4c0a87c1cc329b0e ("Fix truncated fgetc return value in main(espeak-ng.c). [Coverity]"). Before that, EOF would be taken as spurious character input... It's then `p_text[ix-1] = 0;` below which spuriously drops the last character. I guess it just never went noticed just because  people usually pass text in the command or as a file.

sthibaul:
I'll fix that upstream and upload fixes in debian

sthibaul:
> I guess it just never went noticed just because people usually pass text in the command or as a file.

Or `echo` with  a newline character

sthibaul:
The fix was accepted for debian stable

cwendling:
@sthibaul thanks a lot for the proper fix and importing it everywhere in Debian!

PR #982, opened 2024-12-24
Fix compilation on OpenBSD.

brad0:
speechd.h:63: error: redefinition of 'union semun'

### DIFF
--- a/src/modules/module_utils.h
+++ b/src/modules/module_utils.h
@@ -36,8 +36,6 @@
 #include <errno.h>
 #include <dotconf.h>
 
-#include <sys/ipc.h>
-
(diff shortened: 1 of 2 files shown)


sthibaul:
Mmm, we don't seem to be using it actually? I'd say just remove the structure definition entirely.

brad0:
> Mmm, we don't seem to be using it actually? I'd say just remove the structure definition entirely.

I didn't know what you meant by that at first. Seeing as the code includes headers for both the older SysV API and POSIX API. I clearly see the POSIX API utilization but not SysV.

sthibaul:
Thanks!

PR #996, opened 2025-01-22
A native output module based on piper TTS.

net-ddavies:
'Native' means it's written in C++, so it does not incur the overhead of the generic output modulebased on the piper command line utility.

### DIFF
--- a/config/modules/cxxpiper.conf
+++ b/config/modules/cxxpiper.conf
@@ -1,0 +1,71 @@
+#
+# Configuration for cxxpiper speech dispatcher output module.
+#
+
+Debug 0
+
+# Piper doesn't have voices ala speech- dispatcher.  Piper has a
+# "model" and the model's "configuration".  A model/config may be
+# single speaker or multi-speaker.  Single speaker models produce a
+# single speech dispatcher voice.  The speech dispatcher voice name
+# can be listed with 'spd-say -o cxxpiper -L', but it is
+# not needed as the voice will be the default, and only, voice available.
+#
+# Piper multi-speaker models produce a discrete speech dispatcher
+# voice for each speaker the model suports.  The configuration
+# file for a multi-speaker model contains a "speaker_id_map" object
+# that lists an integer speaker id and string mneumonic for each
+# speaker supported by the model.  Since speech dispatcher has no
+# notion of speaker id, speaker selection details are hidden from the
+# user by instead exposing voices of the form
+# <model-name>~<speaker-id>~mneumonic with the output module
+# mapping between "voice names" and the current model's speakers.
+
+# NB:  Unsure if onnx models may allow different " languages within
+# the same multi-speaker model.  REgardless, if there's sufficient memory: it might be possible to load multiple
+# models and have cxxpiper select between them, while presenting the union of
+# the speakers and languages of each model.
+# For now all directives with language fields require the language code, but ignore it.
+
+# ModelPath and ConfigPath are required.  There should be exactly one of each of them.
+ModelPath "/usr/share/piper/voices/clean100.onnx"
+ConfigPath "/usr/share/piper/voices/clean100.onnx.json"
+
+# For single-speaker models, DefaultVoice is ignored, and logged as such, with
+# a warning.  For multi-speaker models, DefaultVoice is optional.  If it is
+# not specified, the first speaker of the multi-speaker model becomes
+# the default speaker for the lifetime of the cxxpiper output module
+# and also for future runtimes unless this configuration is changed.
+# When specified along with a multi-speaker model, the argument is a string that matches one of the
+# "voices" listed by spd-say -o cxxpiper -L .  Note that piper's
+# notion of "speaker" appears to the user as the "voice" concept of
+# speech dispatcher.  This is pretty much invisible to the user, but
+# note that it means that the strings listed in the .json
+# configuration file in the speaker_id_map object are not the same as
+# the voices listed by spd-say (i.e. the speaker ids are substrings of
+# the listed voices).  We could also match the substrings, but we
+# don't at least for now, only the "voice" string string is
+# recognized.  If the voice string can't be matched against the voices
+# found when the model is loaded, then the first speaker becomes the
+# default for the lifetime of the output module.  A warning is logged
+# if the string can't be matched and the voice name of the first
+# speaker (index 0) is included in the warning message.
+DefaultVoice "clean100~2~5393"
+
+# AddVoice (optional) reused from the generic output module.  This maps types to voice names within a language code.
+# It does not do anything useful for single speaker models and is ignored.  For
+# multi-speaker models the language code is required, but ignored, at least for now.
+AddVoice "en_US" "MALE1" "clean100~33~8419"
+AddVoice "en_US" "FEMALE1" "clean100~25~4137"
+
+# Sound Icons are configured and work like espeak.
+SoundIconFolder "/usr/share/sounds/sound-icons/"
+SoundIconVolume 0
+
+# Piper uses ESpeak NG sometimes, depending on the model.  Piper distributes this and distros
+# may provide it.  Default is "/usr/share/espeak-ng-data/".
+# It should probably be considered required, but if a model doesn't use espeak
+# it might work to omit it.
+ESpeakNGDataDirPath "/usr/share/espeak-ng-data/"
+
+# End of cxxpiper.conf
--- a/src/modules/Makefile.am
+++ b/src/modules/Makefile.am
@@ -27,7 +27,6 @@ EXTRA_DIST += dummy-message.txt
 CLEANFILES = dummy-message.wav
 
 inc_local = -I$(top_srcdir)/include -I$(top_srcdir)/src/common
-LDFLAGS =
 
 if DARWIN_HOST
 LDFLAGS += -Wl,-U,_module_speak -Wl,-U,_module_speak_sync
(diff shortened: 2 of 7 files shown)


config/modules/cxxpiper.conf
```diff
@@ -0,0 +1,114 @@
...
+#        english-us+klatt4                    en-US                   klatt4
+#         english-us+male2                    en-US                    male2
+#         english-us+male3                    en-US                    male3
+#         english-us+male4                    en-US                    male4
+#         english-us+male5                    en-US                    male5
+#         english-us+male6                    en-US                    male6
+#         english-us+male7                    en-US                    male7
+#       english-us+whisper                    en-US                  whisper
+#       english-us+female1                    en-US                  female1
+##         english-us+croak                    en-US                    croak
```
sthibaul, line 95: what are these lines?

config/modules/cxxpiper.conf
```diff
@@ -0,0 +1,114 @@
...
+# These are optional.  'None' and 'All' are constant -- the former meaning as much omission and the latter
+# meaning as little, of punctuation, as possible.
+PunctSome "-_+*/\\&$%^~!@#"
+PunctMost "(){}[]<>;:-_+*/\\&$%^~!@#\"'"
+
+# Piper uses ESpeak NG sometimes, depending on the model.  Piper distributes this and distros
+# may provide it.  Default is "/usr/share/espeak-ng-data/".
+# It should probably be considered required, but if a model doesn't use espeak
+# it might work to omit it.
+ESpeakNGDataDirPath "/home/ddavies/src/piper/install/espeak-ng-data"
```
sthibaul, line 112: The default should be fine, so leave it as such in the configuration file.

src/modules/Makefile.am
```diff
@@ -287,6 +287,25 @@ CLEANFILES += $(EXTRA_sd_kali_DEPENDENCIES)
...
+	-I$(PIPER_SRC_DIR)/src/cpp/ \
+	$(AM_CPPFLAGS)
+sd_cxxpiper_LDADD = $(top_builddir)/src/common/libcommon.la \
+	-L$(PIPER_LIB_DIR) \
+	-lpiper_phonemize \
+	-lonnxruntime \
+	-lespeak-ng \
+	-lrubberband \
+	$(SNDFILE_LIBS) \
+	$(common_LDADD)
```
sthibaul, line 307: We really need to make this optional, because at lot of people will not have piper installed on their system. Actually pipe don't seem to have a real installation makefile target that would properly install header files? This really should be fixed by upstream, otherwise it'll be a mess...

This is normally done in `configure.ac`, you can see examples, see for instance `check for svox pico support` which checks for the presence of the `libttspico` library before enabling the `pico_support` condition, which is then used in `Makefile.am`.

src/modules/cxxpiper.cpp
```diff
@@ -0,0 +1,1223 @@
...
+#include <mutex>
+#include <stdexcept>
+#include <vector>
+
+#include <espeak-ng/speak_lib.h>
+#include <onnxruntime_cxx_api.h>
+#include "json.hpp"
+#include "piper.hpp"
+#include "utf8.h"
+#include "wavfile.hpp"
```
sthibaul, line 20: These should really be `<foo.h>` instead, `"foo.h"` is only for the case when the file is in the same directory

src/modules/cxxpiper.cpp
```diff
@@ -0,0 +1,1223 @@
+
```
sthibaul, line 1: This needs to get a header providing the licence under which you ship the module. Apparently you have copy/pasted a lot of code from piper, so it *has* to respect the licensing, so copy/paste the content of LICENSE.md from there.

src/modules/cxxpiper.cpp
```diff
@@ -0,0 +1,1223 @@
...
+#include <stdexcept>
+#include <vector>
+
+#include <espeak-ng/speak_lib.h>
+#include <onnxruntime_cxx_api.h>
+#include "json.hpp"
+#include "piper.hpp"
+#include "utf8.h"
+#include "wavfile.hpp"
+#include "../rubberband/RubberBandStretcher.h"
```
sthibaul, line 21: A `-I/where/rubberband/is` should be added to `Makefile.am` instead. It's available as system-provided packages in Debian for instance.

src/modules/cxxpiper.cpp
```diff
@@ -0,0 +1,1223 @@
...
+// Copied from piper distribution.
+// Get the first UTF-8 codepoint of a string
+    Phoneme getCodepoint(std::string s)
+    {
+	utf8::iterator character_iter(s.begin(), s.begin(), s.end());
+	return *character_iter;
+    }
+
+// Copied from piper distribution.
+    void parsePhonemizeConfig(json &configRoot, PhonemizeConfig &phonemizeConfig)
```
sthibaul, line 90: It's quite sad that piper doesn't provide these as library functions. It'd be good to tell upstream to do so, rather than copying their code

src/modules/cxxpiper.cpp
```diff
@@ -0,0 +1,1223 @@
...
+    std::vector<int16_t> audioBuffer;
+    textToAudio(config, voice, text, audioBuffer, result, NULL);
+    auto synthesisConfig = voice.synthesisConfig;
+    writeWavHeader(synthesisConfig.sampleRate, synthesisConfig.sampleWidth,
+		   synthesisConfig.channels, (int32_t)audioBuffer.size(),
+		   audioFile);
+    audioFile.write((const char *)audioBuffer.data(),
+		    sizeof(int16_t) * audioBuffer.size());
+}
+#endif
```
sthibaul, line 550: We won't need it indeed, you can drop it

src/modules/cxxpiper.cpp
```diff
@@ -0,0 +1,1223 @@
...
+    static SPDVoice **cxxpiper_allocate_voice_list();
+    static void cxxpiper_free_voice_list();
+    static void cxxpiper_handle_text(const char *);
+    static void cxxpiper_handle_sound_icon(const char *);
+    static void cxxpiper_set_language(char *);
+    static void cxxpiper_set_synthesis_voice(char *voice_name);
+    static void cxxpiper_set_voice_type(SPDVoiceType);
+  static int cxxpiper_voice_name_to_speaker_id(const char *);
+
+    MOD_OPTION_1_INT(UseCUDA)
```
sthibaul, line 713: Does CUDA already work?

src/modules/cxxpiper.cpp
```diff
@@ -0,0 +1,1223 @@
...
+	runConfig.modelPath = filesystem::path(ModelPath);
+	runConfig.modelConfigPath = filesystem::path(ConfigPath);
+	runConfig.eSpeakDataPath = filesystem::path(ESpeakNGDataDirPath);
+	runConfig.outputPath = nullopt;
+	char *default_voice = module_getdefaultvoice();
+	runConfig.defaultVoiceName = default_voice;
+	runConfig.speakerId = cxxpiper_voice_name_to_speaker_id(runConfig.defaultVoiceName);
+	DBG("Default Voice is %s", runConfig.defaultVoiceName);
+	cxxpiper::loadVoice(piperConfig, runConfig.modelPath.string(), runConfig.modelConfigPath.string(),
+				voice, runConfig.speakerId, runConfig.useCuda);
```
sthibaul, line 757: We really need to catch errors here, and return -1 on failure, to make sure that the voice doesn't show up in orca's list and users accidently enabling a voice that actually doesn't work and then they're doomed

src/modules/cxxpiper.cpp
```diff
@@ -0,0 +1,1223 @@
...
+	    if (voice.phonemizeConfig.phonemeType == piper::eSpeakPhonemes) {
+		piperConfig.eSpeakDataPath = runConfig.eSpeakDataPath.value().string();
+	    }
+	    else {
+		piperConfig.useESpeak = false;
+	    }
+	    cxxpiper::initialize(piperConfig);
```
sthibaul, line 764: Same here

src/modules/cxxpiper.cpp
```diff
@@ -0,0 +1,1223 @@
...
+	    cxxpiper_voice_list = cxxpiper_allocate_voice_list();
+	    cxxpiper_alloc_cbuf();
+	    *status_info = g_strdup(DBG_MODNAME " Initialized successfully.");
+	    return 0;
+	    }
+
+	    SPDVoice **module_list_voices(void)
+    {
```
sthibaul, line 772: There's odd indentation, it is not consistent

src/modules/cxxpiper.cpp
```diff
@@ -0,0 +1,1223 @@
...
+	    break;
+	case SPD_MSGTYPE_TEXT: {
+	    cxxpiper_handle_text(data);
+	    break;
+	}
+	case SPD_MSGTYPE_SOUND_ICON: {
+	    cxxpiper_handle_sound_icon(data);
+	    break;
+	}
+	}
```
sthibaul, line 796: This double `}` is odd, better either indent the whole content of the `switch`, or the content of each `case`.

src/modules/cxxpiper.cpp
```diff
@@ -0,0 +1,1223 @@
...
+#endif
+	    audioReady = true;
+	    ++cbCnt;
+	    cbTot += audioBuffer.size();	    //cerr << "CVaudio notivy 1" << endl;
+	    cvAudio.notify_one();
+	};
+	DBG("Sending begin event");
+	module_report_event_begin();
+	(void)cxxpiper::textToAudio(piperConfig, voice, cmdInp,
+				    audioBuffer, result, audioCallback);
```
sthibaul, line 1035: Does piper support pipelining? To start sending audio to the server before the whole audio is produced. That would be really important to get good reactivity (but it's fine not to have it for first inclusion)

src/modules/cxxpiper.cpp
```diff
@@ -0,0 +1,1223 @@
...
+    {
+	stop_requested = 0;
+	UPDATE_STRING_PARAMETER(voice.language, cxxpiper_set_language);
+	UPDATE_PARAMETER(voice_type, cxxpiper_set_voice_type);
+	UPDATE_STRING_PARAMETER(voice.name, cxxpiper_set_synthesis_voice);
+	module_speak_ok();
+	switch (msgtype) {
+	case SPD_MSGTYPE_CHAR:
+	case SPD_MSGTYPE_KEY:
+	case SPD_MSGTYPE_SPELL:
```
sthibaul, line 786: We shouldn't ignore them, that'd break some screen reading support.

sthibaul:
> We'll also need a CI to make sure it works, but again I can work on that.

I have added to CI a rule to install piper in `/opt`

net-ddavies:
Grateful for your fixes to CI and Makefile.am -- thank you.

src/modules/cxxpiper.cpp
sthibaul: This is really a must

src/modules/cxxpiper.cpp
sthibaul: This is really needed.

Issue #866, opened 2023-11-06
module request: piper


KiaraGrouwstra:
[`piper`](https://github.com/rhasspy/piper) is 'a fast, local neural text to speech system' ([samples here](https://rhasspy.github.io/piper-samples/)).
it would be nice to have `speechd` support this as well.

cross-post: https://github.com/rhasspy/piper/issues/265

/cc @Elleo who has done some work integrating these thru [pied](https://github.com/Elleo/pied/).

csukuangfj:
Also cc @Elleo . You may find [sherpa-onnx](https://github.com/k2-fsa/sherpa-onnx) interesting.
It supports both speech-to-text and text-to-speech.

Elleo:
Just for a little context on what I'm doing, Pied can currently configure speech dispatcher to work with Piper through the sd_generic module, but my long term plan is to create a piper speech dispatcher module that can be kept loaded to further reduce latency and add support for speed/pitch/etc. changes

@csukuangfj Thanks, that's interesting, I'll check it out!

coderalpha:
I'm trying to integrate Piper through the sd_generic module. But I get the error:
` speechd:  Error: Module reported error in request from speechd (code 3xx): 300-Opening sound device failed. Reason: Cannot open plugin server. error: file not found.`

I cant find any information on this error. Any help will be appreciated!

I added the module in speechd.conf:
`AddModule "piper"                    "sd_generic" "piper.conf"'`
`DefaultVoiceType "FEMALE1"`
`DefaultModule "piper"`
`DefaultLanguage "en"`
`AudioOutputMethod "libao"`

And created the piper.conf file in the /etc/speech-dispatcher/modules directory:
`AddVoice "en" "FEMALE1" "en_US-amy-medium.onnx"`
`DefaultVoice "en_US-amy-medium.onnx"`
`GenericExecuteSynth "echo \'$DATA\' | /home/dev/Apps/piper/piper --model /home/dev/Apps/piper/models/en_US-amy-medium.onnx --output_raw | paplay"`

sthibaul:
As mentioned in the issue template, add Debug 1 to the speechd config file and the module config file, and get the corresponding log files, so we get to know what exactly went wrong.

sthibaul:
(of course, the issue template knows better, that's why we write documentation, so we don't have to rely on our memory: it's LogLevel 5 in the speechd config, and indeed Debug 1 in the module config)

coderalpha:
[...]

PR #998, opened 2025-01-24
sd_piper: add module for piper speech synthesis

samoverton:
I have been working on a dedicated piper module (for #866). I just pulled from master and noticed that the latest commit (bec5519) alludes to work in progress on a cxxpiper module, so I wanted to share my work asap to ensure that I wasn't stepping on any toes and see if you would like to collaborate.

The module makes use of the user's installation of the piper binary instead of linking in the piper codebase directly. This means that each speak request forks a child process with the appropriate arguments. Communication with the child is done over pipes. Server audio is used so speechd handles output to the configured audio device.

The module supports usual speechd configuration and runtime parameters:
* `AddVoice` directives for configuring the voice models to use
* `rate`- setting of voice speed (between 0.5x and 2x)
* TEXT, CHAR, SPELL message types
* STOP and PAUSE events

As well as some piper specific configuration:
* Voice sample rate is read from model manifest json file
* `sentence_silence` - seconds of silence after each sentence
* `noise_scale` - generator noise
* `noise_w` - phoneme width noise
* audio buffer size (ms)

This changeset is based on master, but I originally did this work on the 0.11 branch, so it can be easily backported if required.

### DIFF
--- a/config/modules/piper.conf
+++ b/config/modules/piper.conf
@@ -1,0 +1,29 @@
+# Debug 1
+
+# Location of the piper binary
+PiperBinPath "/usr/lib/piper/piper"
+
+# Location of voice model files
+PiperModelsPath "/usr/share/piper-voices"
+
+# The amount of voice data (ms) to buffer from piper before returning it to
+# speech-dispatcher for playback
+#   default 200
+PiperAudioBufferMsec 200
+
+# For the following parameters, value/100 is passed to piper cmdline
+
+# seconds of silence after each sentence
+#   default 20 -> 0.20 sec
+PiperSentenceSilence 20
+
+# generator noise
+#   default 67 -> 0.67
+PiperNoiseScale 67
+
+# phoneme width noise
+#   default 80 -> 0.80
+PiperNoiseWidth 80
+
+# Voice definitions. The name must be the filename within ${PiperModelsPath}/
+AddVoice "en-gb" "female1" "en_GB-alba-medium.onnx"
--- a/config/speechd.conf
+++ b/config/speechd.conf
@@ -288,6 +288,7 @@ SymbolsPreprocFile "symbols-fallback.dic"
 #AddModule "baratinoo"                "sd_baratinoo" "baratinoo.conf"
 #AddModule "rhvoice"                  "sd_rhvoice"   "rhvoice.conf"
 #AddModule "voxin"                    "sd_voxin"     "voxin.conf"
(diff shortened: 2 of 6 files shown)


samoverton:
I just came across #996 and I see that it takes a different approach, but I will leave this work here in case it is useful to anyone, or in the event that bringing piper code into the speech-dispatcher codebase turns out not to be viable.

azakharchenko-msol:
@samoverton Great work :+1:  works great and solves #999 
Not exactly understand the logic of `const char* piper_get_voice(SPDMsgSettings* p_settings)`
the spd-say works only if I pass `-t female1` otherwise it prints "no voice found" in log.
Is the default voice should be set somewhere?

samoverton:
> the spd-say works only if I pass `-t female1` otherwise it prints "no voice found" in log. Is the default voice should be set somewhere?

You can set the default voice for the module in the `piper.conf`:
```
DefaultVoice "en_GB-alba-medium.onnx"
```

Or you can set the default language and voice-type in `speechd.conf`:
```
DefaultVoiceType "female1"
DefaultLanguage "en-GB"
```

You're right that there is some weird behavior here that's worth looking at though. Did you have any of these defaults set in `speechd.conf`?

azakharchenko-msol:
@samoverton Thank you, I missed `DefaultVoiceType "female1"` in my `speechd.conf`

net-ddavies:
Very happy to collaborate with you on piper work.  Your work got me started so big Thank You!!  Off the top of my head it would be great to unify the .conf as much as we can.

Samuel T mentions sttreaming/pipelining in his review of cxxpiper and
I'm very much thinking of that.  It would be very cool to have real
time voice adjustment for any parameter.  It would actually be useful
and not just cool when tweaking voices by ear to get them just right.
I've done quite a bit of experimenting with this and ran into trouble
so I shelved it for the moment, but it is a live issue.

I was also thinking of caching and how it might be nice to have a
caching module that all the OM's could share, but that's not piper related, of course.

What are your thoughts on how to work together on stuff?

Thanks!
Derek

sthibaul:
Hello,
#996 was indeed a more promising approach, because it allows to keep the model in memory and avoid reloading it each time. I was able to fix the issue I was seeing with it, it now works for me. It will be very useful to extend it to support loading several models at a time (and ideally, auto-detect them from /usr/share/piper/voices), #1092 can be used to track these.

PR #1007, opened 2025-02-28
Mixed case words

### LINKED ISSUE #946: Pre-process mixed-case words for synthesizers that don't handle them correctly
I recently removed the pre-processing of mixed-case words (e.g. "HelloWorld") from Orca because from my testing the synthesizers were already doing this work, so there was no need for Orca to do it.

End user testing has revealed that there are a few which do not, such as voxin with IBMTTS. Therefore, would it be possible for speech-dispatcher to do this work at least in the case of no-longer-maintained synthesizers?

raphaelpoitevin:
fixes #946

### DIFF
--- a/src/api/python/speechd_config/.gitignore
+++ b/src/api/python/speechd_config/.gitignore
@@ -20,5 +20,6 @@
 /TAGS
 /buildconfig.py
 /so_locations
+/spd-conf.1
 /speechd.desktop
 /tags
--- a/src/api/python/speechd_config/.gitignore
+++ b/src/api/python/speechd_config/.gitignore
@@ -20,6 +20,5 @@
 /TAGS
 /buildconfig.py
 /so_locations
-/spd-conf.1
 /speechd.desktop
 /tags
--- a/src/modules/module_utils.c
+++ b/src/modules/module_utils.c
@@ -49,10 +49,14 @@ const char *module_name;
 int current_index_mark;
 char *module_index_mark;
 
+typedef struct {
+	gchar *pattern;
+	gchar *replace;
+} MulticasesString;
+
 gchar *module_multicases_string_replace(
 	const gchar *text,
-	const gchar *pattern,
-	const gchar *replace);
+	const MulticasesString *mcstr);
 
 void MSG(int level, const char *format, ...) {
 	if (level < 4 || Debug) {
@@ -420,22 +424,21 @@ void module_strip_punctuation_default(char *buf)
 
 gchar *module_multicases_string_replace(
 	const gchar *text,
-	const gchar *pattern,
-	const gchar *replace)
+	const MulticasesString *mcstr)
 {
 	GRegex *regex;
 	GError *error = NULL;
 	gchar *result;
 
 	result = g_strdup(text);
-	regex = g_regex_new(pattern, G_REGEX_OPTIMIZE, 0, &error);
+	regex = g_regex_new(mcstr->pattern, G_REGEX_OPTIMIZE, 0, &error);
 	if (!regex) {
 		DBG("ERROR compiling regular expression: %s.", error->message);
 		g_error_free(error);
 		return result;
 	}
 
-	result = g_regex_replace(regex, text, -1, 0, replace, G_REGEX_MATCH_DEFAULT, NULL);
+	result = g_regex_replace(regex, text, -1, 0, mcstr->replace, G_REGEX_MATCH_DEFAULT, NULL);
 
 	g_error_free(error);
 	g_regex_unref(regex);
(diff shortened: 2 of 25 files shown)


src/modules/module_utils.c
```diff
@@ -413,6 +418,62 @@ void module_strip_punctuation_default(char *buf)
...
+{
+	gchar *pattern[4];
+	gchar *replace[4];
+	gchar *multicases;
+	gchar *text;
+	guint i;
+
+	assert(message != NULL);
+
+	pattern[0] = "([a-z]+)([A-Z][a-z]+)";
```
sthibaul, line 456: you probably want to add word boundaries at beginnings and ends of the patterns? (`\b`) to avoid matching in unexpected ways

src/modules/module_utils.c
```diff
@@ -413,6 +418,62 @@ void module_strip_punctuation_default(char *buf)
...
+
+	g_error_free(error);
+	g_regex_unref(regex);
+							 
+	return result;
+}
+
+char *module_multicases_string(const char *message)
+{
+	gchar *pattern[4];
```
sthibaul, line 448: you can make these completely static:

```
static gchar *pattern[] = {
   "...",
   "...",
};
```

Ideally, even define a small structure that contains the pattern and the replace, and use an array of such structures

src/modules/module_utils.c
```diff
@@ -413,6 +418,62 @@ void module_strip_punctuation_default(char *buf)
...
+	pattern[2] = "([A-Z]{2}[A-Z]+)([a-z]+)";
+	replace[2] = "\\1 \\2";
+	pattern[3] = "([A-Z])([A-Z][a-z]+)";
+	replace[3] = "\\1 \\2";
+
+	text = g_strdup(message);
+	
+	for (i = 0; i < 4; i++) {
+		multicases = module_multicases_string_replace(text, pattern[i], replace[i]);
+		text = g_strdup(multicases);
```
sthibaul, line 469: take care of memory leaks, you are duplicating the strings a lot without ever freeing them

src/server/.gitignore
```diff
@@ -22,6 +22,5 @@
 /TAGS
 /so_locations
 /speech-dispatcher
-/speech-dispatcher.1
```
sthibaul, line 25: spurious change :)

src/api/python/speechd_config/.gitignore
```diff
@@ -20,6 +20,5 @@
 /TAGS
 /buildconfig.py
 /so_locations
-/spd-conf.1
```
sthibaul, line 23: Spurious change :)

sthibaul (changes requested):
It looks good overall! Just a few comments, and the must-fix is really the memory leaks

src/modules/module_utils.c
```diff
@@ -445,33 +448,26 @@ gchar *module_multicases_string_replace(
 
 char *module_multicases_string(const char *message)
 {
-	gchar *pattern[4];
-	gchar *replace[4];
+	static MulticasesString mcstr[] = {
+		{"(\\B[a-z]+\\B)(\\B[A-Z][a-z]+\\B)", "\\1 \\2"},
```
sthibaul, line 452: `\\B` ? why not `\\b`? `\\B` is non-boundary, while you do want a boundary at begin/end of your pattern, i.e. rather

`"\\b([a-z]+)([A-Z][a-z]+)\\b"`

src/modules/module_utils.c
```diff
@@ -413,6 +422,54 @@ void module_strip_punctuation_default(char *buf)
 	module_strip_punctuation_some(buf, "~#$%^&*+=|<>[]_");
 }
 
+gchar *module_multicases_string_replace(
+	const gchar *text,
```
sthibaul, line 426: `text` is never freed. The caller expects `module_multicases_string_replace` to free it (and that should be documented so)

src/modules/module_utils.c
```diff
@@ -413,6 +422,54 @@ void module_strip_punctuation_default(char *buf)
...
+	const MulticasesString *mcstr)
+{
+	GRegex *regex;
+	GError *error = NULL;
+	gchar *result;
+
+	result = g_strdup(text);
```
sthibaul, line 433: You don't need to `strdup` `text`. In the error case, you can just return `text`. In the success case, you will return the result.

src/modules/module_utils.c
raphaelpoitevin: On 3/4/25 16:33, Samuel Thibault wrote:
> @sthibaul commented on this pull request.
>
>
>
>> @@ -445,33 +448,26 @@ gchar *module_multicases_string_replace(
>     char *module_multicases_string(const char *message)
>   {
> -    gchar *pattern[4];
> -    gchar *replace[4];
> +    static MulticasesString mcstr[] = {
> +        {"(\\B[a-z]+\\B)(\\B[A-Z][a-z]+\\B)", "\\1 \\2"},
>
> `\\B` ? why not `\\b`? `\\B` is non-boundary, while you do want a boundary at begin/end of your pattern, i.e. rather

When I tried I had some side-effects and I misunderstood that page https://docs.gtk.org/glib/struct.Regex.html which made me confused. Sorry.
>
> `"\\b([a-z]+)([A-Z][a-z]+)\\b"`

I tried but it seems to break attempted effects.
For instance: "HelloEvryBody" works without boundaries words, but nomore with. I certainly missed something.

-- 
Raphaël

raphaelpoitevin: For the moment I didn’t yet correct boundaries words.

sthibaul: Ah, you really want the replacement to happen everywhere, not just as whole words, I had misunderstood that. Then you don't need any `\\b` or `\B`.

src/modules/module_utils.c
```diff
@@ -413,6 +422,50 @@ void module_strip_punctuation_default(char *buf)
...
+	gchar *result;
+
+	regex = g_regex_new(mcstr->pattern, G_REGEX_OPTIMIZE, 0, &error);
+	if (!regex) {
+		DBG("ERROR compiling regular expression: %s.", error->message);
+		g_error_free(error);
+		return text;
+	}
+
+	result = g_regex_replace(regex, text, -1, 0, mcstr->replace, G_REGEX_MATCH_DEFAULT, NULL);
```
sthibaul, line 440: you want to look for errors. On error, just log it and return `text` as it is. Without error, you need to free `text`

src/modules/module_utils.c
```diff
@@ -49,6 +49,15 @@ const char *module_name;
...
 char *module_index_mark;
 
+typedef struct {
+	gchar *pattern;
+	gchar *replace;
+} MulticasesString;
+
+gchar *module_multicases_string_replace(
+	gchar *text,
+	const MulticasesString *mcstr);
```
sthibaul, line 59: you don't need this prototype declaration

src/modules/module_utils.c
```diff
@@ -413,6 +422,50 @@ void module_strip_punctuation_default(char *buf)
 	module_strip_punctuation_some(buf, "~#$%^&*+=|<>[]_");
 }
 
+gchar *module_multicases_string_replace(
```
sthibaul, line 425: you can make it `static`

sthibaul:
Thanks!

PR #1008, opened 2025-03-01
doc/cs: Fix mangled encoding

jtojnar:
f0a0707628767c02719adbeae6d921214960e368 only intended to fix “precission”
typo at the end of file. Unfortunately, the document was opened as UTF-8
encoding instead of the actual ISO-8859-2, resulting in all non-ASCII
characters being replaced with U+FFFD Replacement Character.

I reverted all the changes from the commit that affected
`doc/speech-dispatcher-cs.texi` except for the typo fix itself.


sthibaul:
(master is now equal to 1b1cd50f36398d78a8307d71cfe6f7f788b0403d)

sthibaul:
sorry I missed your update (apparently github doesn't notify of rebase?)
thanks!

PR #1050, opened 2025-09-21
symbols: Also report successes

sthibaul:
So people do not care too much about the failures.


shindere:
Yes, that's a super good idea.

In the same vein, it would be super helpful to rephrase the messages
that report a successful read, as they use the "error" keywoard
which makes finding real, genuine errors harder. Perhaps messages
indicating a successful complete read could also be removed...

sthibaul:
> rephrase the messages that report a successful read, as they use the "error" keywoard

I don't see which message you refer to?

> Perhaps messages indicating a successful complete read could also be removed...

Also, which message do you refer to?

PR #1063, opened 2025-11-11
Update CLDR to version 48, Orca and NVDA symbols to latest version

Nardol:


locale/bs/orca-math.dic
```diff
@@ -7,73 +7,36 @@
...
-∢	sferni ugao	none
-∤	nije djeljivo	none
-∦	nije paralelno	none
-∨	logičko ILI	none
-∭	trostruki integral	none
-∯	povtšinski integral	none
-∰	zapreminski integral	none
-∱	integral u pravcu kazaljke	none
-∲	konturni integral u pravcu kazaljke	none
-∳	konturni integral obrnuto od pravca kazaljke	none
```
sthibaul: and these

sthibaul (changes requested):
I wonder how it is that these are getting removed in the script output?

Nardol:
I made a debug log for emojis because I don't understand the issue about emojis not pronounced at all.

[kali.log](https://github.com/user-attachments/files/23525592/kali.log)
[speech-dispatcher.log](https://github.com/user-attachments/files/23525594/speech-dispatcher.log)

I might add removed symbols from Orca again manually because I don't know what happened wrong when importing...

locale/bs/orca-math.dic
sthibaul: And same for this one

sthibaul:
> I made a debug log for emojis because I don't understand the issue about emojis not pronounced at all.
> 
> [kali.log](https://github.com/user-attachments/files/23525592/kali.log) [speech-dispatcher.log](https://github.com/user-attachments/files/23525594/speech-dispatcher.log)
> 
> I might add removed symbols from Orca again manually because I don't know what happened wrong when importing...

In the log we can read `ERROR compiling regular expression: Erreur à la compilation de l’expression régulière`

sthibaul:
`au caractère 15637 : l’expression régulière est trop grande. This is likely due to an invalid complex symbol regular expression in locale fr.`

sthibaul:
I have implemented in master a split of the table in chunks of 1000 symbols, so we avoid hitting pcre limitations. Could you rebase on it, and check that this works fine for you?

Nardol:
Emojis work but I notice a strangeness with the "'" symbol which is treated as a space by Orca in the case of a message E.G. "Lecteur d'écran activé" or when specifying "niveau d'arborescence 1" in Caja.
Some logs:
[kali.log](https://github.com/user-attachments/files/23579575/kali.log)
[speech-dispatcher.log](https://github.com/user-attachments/files/23579576/speech-dispatcher.log)
I also tested on master.

PR #1068, opened 2025-12-17
symbols: Better describe the files being loaded

### LINKED ISSUE #1049: speechd can't find its symbols preprocessing files when run system-wide
Its logs show messages like

```
speechd: Failed to open file '/var/run/speech-dispatcher/.config/speech-dispatcher/locale/fr/gender-neutral.dic': No such file or directory
```

I couldn't find an obvious way to make the seearched paths adjust the real ones. I tried creating a `.config` directory in `/run/speech-dispatcher` and, in this directory, a `locale` symlink to `/usr/share/speech-dispatcher/locale` but this does not persist after reboot and feels hackish anyway.

One other possibility could be to create that `locale` symlink in `/etc/speech-dispatcher` but I am not yet convinced this would be the right way to fix this issue.

sthibaul:
Printing all errors is not productive, better document which paths are being looked into, and the files actually getting loaded.

Fixes #1049

### DIFF
--- a/src/server/symbols.c
+++ b/src/server/symbols.c
@@ -662,7 +662,11 @@ static int speech_symbols_load(SpeechSymbols *ss, const char *filename, gboolean
 
 	fp = fopen(filename, "r");
 	if (!fp) {
-		MSG2(1, "symbols", "Failed to open file '%s': %s", filename, g_strerror(errno));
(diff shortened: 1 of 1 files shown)


shindere:
Many thanks for that!

PR #1080, opened 2026-02-22
Remove REGEX dependency from Common Lisp ASDF:defsystem definition.

mrd:
The REGEX library is not actually used in this Lisp code, and the REGEX library itself is deprecated, and has not been supported for many years or decades by its author.


sthibaul:
Thanks!

PR #1086, opened 2026-04-12
Add option to build only libspeechd and not the rest of the library

ethindp:
So, this PR got turned into a mess when I did it against 0.12.1. Oops!

This was what I had to do to get libspeechd to build in vcpkg. It would be nice if it was a part of speech dispatcher itself so I didn't need to keep maintaining what is a very fragile patch indeed. Sorry about all the extraneous commits that got included (thank you, Git).


sthibaul:
I cannot take the time to clean this up. Please rebase onto master.

sthibaul:
Fortunately, the actual useful commit was easy to find in the series,  I applied it by hand

ethindp:
I'm sorry for making a mess. Although I'm uncertain why a squash commit wouldn't have worked here. I'm also not super familiar with autotools so it's possible I might've been able to done things in some better way than how I did.

sthibaul:
> I'm uncertain why a squash commit wouldn't have worked here

The question wasn't really about what to commit (git would work indeed), but being able to review what is to be commited. See the "Files changed" tab, it's full of unrelated changes.

ethindp:
@sthibaul Oh yeah that happens to me sometimes and I'm not entirely sure why. It's a weird quirk of Git I've just kinda gotten used to

sthibaul:
This is really a matter of rebasing onto the targetted branch.

PR #1103, opened 2026-06-13
Fix SSML stripping inside tags

DaTeVo:
Fix `module_strip_ssml()` to ignore XML entities inside SSML tags.

Previously:

```xml
<tag attr="Tom &amp; Jerry">Hello</tag>
```

produced:

```text
&Hello
```

It now produces:

```text
Hello
```

### DIFF
--- a/src/modules/module_utils.c
+++ b/src/modules/module_utils.c
@@ -389,6 +389,8 @@ char *module_strip_ssml(const char *message)
 			omit = 0;
 			continue;
 		}
+		if (omit && i != len)
+			continue;
 		if (!strncmp(&(message[i]), "&lt;", 4)) {
 			i += 3;
 			out[n++] = '<';
--- a/src/modules/module_utils.c
+++ b/src/modules/module_utils.c
@@ -379,7 +379,7 @@ char *module_strip_ssml(const char *message)
 	len = strlen(message);
 	out = (char *)g_malloc(sizeof(char) * (len + 1));
 
-	for (i = 0, n = 0; i <= len; i++) {
+	for (i = 0, n = 0; i < len; i++) {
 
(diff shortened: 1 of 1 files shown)


sthibaul:
Thanks for spotting this!

I'm thinking that the loop could instead be reworked a bit. I was actually surprised that it was including i == len. We don't actually need that, but we need to terminate the out string. It seems to me we should rather simplify the loop:

- use `i<len` in the for loop,
- add out[n++] = 0 after the loop to terminate the out string,
- and then we can just introduce `if (omit) continue;` where you added your if, and remove the last if, to just copy one character when it's not a known ssml entity.

DaTeVo:
Thank you for the review. I will take care of it on Wednesday.

src/modules/module_utils.c
```diff
@@ -404,9 +406,13 @@ char *module_strip_ssml(const char *message)
 		} else if (!strncmp(&(message[i]), "&apos;", 6)) {
 			i += 5;
 			out[n++] = '\'';
-		} else if (!omit || i == len)
+		} else {
 			out[n++] = message[i];
+		}
 	}
+
+	out[n++] = messaege[i];
```
sthibaul: that won't compile ;)

daxmawal:
```suggestion
	out[n++] = 0;
```

sthibaul:
squashed and pushed, thanks!

daxmawal:
Salut Samuel,

J'aurai 1-2 semaines de retard pour la prochaine PR pour étendre les
modules/fonctions en Python.
Mais, ça avance.

Bonne soirée,

Le jeu. 18 juin 2026 à 01:27, Samuel Thibault ***@***.***> a
écrit :

PR #1104, opened 2026-06-19
Add initial Python module helpers

daxmawal:
This is an initial PR to check whether maintainers are happy with the proposed Python module helper integration before continuing.

It adds a small `speechd_python_modules` Python package. These helper modules are installed into the configured Python installation under:
`$(pythondir)/speechd_python_modules`

Follow-up PRs would add, in order:
- Python equivalents for more C module helper functions/modules needed by Python output modules
- integration of the neural-network [Pocket TTS](https://github.com/kyutai-labs/pocket-tts) module into speechd
- pause/stop support, finish to translate C functions/modules into Python, and further optimizations

Disclosure: Codex assisted me with Autotools, Autoconf, and review.


src/modules/Makefile.am
```diff
@@ -26,6 +26,14 @@ EXTRA_DIST += dummy-message-default.wav
...
 CLEANFILES = dummy-message.wav
 
+if HAVE_PYTHON
+speechd_python_modules_pythondir = $(pythondir)/speechd_python_modules
+dist_speechd_python_modules_python_PYTHON = \
+	speechd_python_modules/__init__.py \
+	speechd_python_modules/module_utils.py \
+	speechd_python_modules/speechd_types.py
+endif
+
```
daxmawal, line 36: Should speechd_python_modules be installed under the existing HAVE_PYTHON condition?

--disable-python is currently documented as disabling Python bindings, but this would also disable Python helpers for output modules.

daxmawal:
I'm not sure if `src/tests/python_module.at` is actually useful for speechd, later, perhaps we could transform this into something equivalent in the latest follow-up PRs to c_api.at once Pocket TTS has been integrated. Or we could just delete it now.

src/modules/Makefile.am
sthibaul: It's fine for HAVE_PYTHON to drive both enabling the client python bindings and the output modules module, we'd just need to update the documentation

src/tests/python_module.at
```diff
@@ -0,0 +1,139 @@
+# python_module.at - python module tests
+#
+# Copyright (C) 2010 Brailcom, o.p.s.
```
sthibaul: I don't think Brailcom has copyright on this content :)

src/modules/speechd_python_modules/__init__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (C) 2001, 2002 Brailcom, o.p.s.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
```
sthibaul: We would rather want to publish the python module under the BSD-2 licence, to make it usable as largely as possible, like `module_main.c` etc.

src/modules/speechd_python_modules/speechd_types.py
```diff
@@ -0,0 +1,45 @@
+#
+# speechd_types.py - types for Speech Dispatcher
+#
+# Copyright (C) 2001, 2002, 2003 Brailcom, o.p.s.
+# Copyright (C) 2004 Brailcom, o.p.s.
+# Copyright (C) 2019-2024 Samuel Thibault <samuel.thibault@ens-lyon.org>
```
sthibaul: I don't think there is anything that still wears these copyrights: interfaces per se (constant names and their values) don't really wear copyrights, and the way you wrote them is pythonish, losing the original C style

src/modules/speechd_python_modules/module_utils.py
```diff
@@ -0,0 +1,72 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+
+def module_strip_ssml(message: str) -> str:
+    nul = message.find("\0")
+    if nul != -1:
+        message = message[:nul]
```
sthibaul: I don't think we will need this in the python version, it's really just C that has to care about \0

src/modules/speechd_python_modules/module_utils.py
```diff
@@ -0,0 +1,72 @@
+#
+# module_utils.py - Module utilities
+#           Functions to help writing output modules for Speech Dispatcher
+# Copyright (C) 2003,2006, 2007 Brailcom, o.p.s.
```
sthibaul: While you reused the algorithm, the writing looks quite different in Python the way you wrote it, so I don't think the copyright still holds.

src/modules/speechd_python_modules/module_utils.py
```diff
@@ -0,0 +1,72 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+
+def module_strip_ssml(message: str) -> str:
```
sthibaul, line 30: I was a bit undecided between this reimplementation of the C function and just adding a binding on top of the existing C function.

But it seems to me that there won't be that much to be leveraged from the C implementation, or else the python bindings will be very C-ish and prevent from having a really pythonistic interface.

So I think it's fine to duplicate such kind of helper, we are not that much performance-sensitive anyway.

sthibaul:
> This is an initial PR to check whether maintainers are happy with the proposed Python module helper integration before continuing.

Thanks! That looks promising.

src/modules/Makefile.am
daxmawal: Updated [here](https://github.com/brailcom/speechd/pull/1104/changes#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R227) with ./configure --help

sthibaul:
thanks!

daxmawal:
I think I'll have finished the next follow up PR within the next 2-3 weeks.

PR #1105, opened 2026-07-26
Add Python output modules

daxmawal:
## Summary

This PR is a follow-up to #1104 and adds reusable helpers for implementing Speech Dispatcher output modules in Python.
Disclosure: Codex assisted with code review and testing.

## Follow-up work

Planned follow-up PRs will:

1. Python output module protocol implementation (`module_main` and `module_process`) and server-side audio transport.
2. add the [Pocket TTS](https://github.com/kyutai-labs/pocket-tts) output module.
3. implement the remaining features, including sound icons, `STOP`, `PAUSE`, and complete custom debug logging.


src/modules/speechd_python_modules/module_utils.py
```diff
@@ -27,6 +27,58 @@
...
+CustomDebugFile = None
+
+
+def module_loglevel_set(cur_item, cur_value):
+    global log_level
+
+    if cur_item != "log_level":
+        return -1
+
+    match = re.match(r"\s*([+-]?[0-9]+)", cur_value)
```
daxmawal, line 44: I didn't find equivalent to strtol()

src/modules/speechd_python_modules/module_readline.py
```diff
@@ -0,0 +1,131 @@
+#
+# module_readline.py - Input buffering for Python output modules.
+#
+# Copyright (C) 2020 Samuel Thibault <samuel.thibault@ens-lyon.org>
```
daxmawal, line 4: The Python code is simply a translation of the C code, it's therefore justified to add his name.

daxmawal:
The Python tests have been successfully run and are available [here](https://github.com/brailcom/speechd/actions/runs/30211926642/job/89819502499?pr=1105#step:13:19).

src/modules/speechd_python_modules/module_readline.py
```diff
@@ -0,0 +1,131 @@
...
+
+_fd_buffers = {}
+
+
+class _ReadBuffer:
+    def __init__(self):
+        self.data = bytearray()
+        self.no_lf = 0
+
+def module_readline(source=None, block=True):
```
sthibaul, line 45: Mmm, module_readline is already available in libspeechd_module, I'd say we can just use it instead of reimplementing it (and thus having to maintain both).

sthibaul: Note that this is different from module_strip_ssml, as module_readline is available under a BSD  license while the existing implementation of module_strip_ssml is available under LGPL.

What I'm thinking is that the python module helper can leverage libspeechd_module to manage the protocol etc. and translate the eventual calls into python calls, plus some helpers to make things easier for module implementors.

src/modules/speechd_python_modules/module_utils.py
```diff
@@ -27,6 +27,58 @@
...
 
+import re
+
+
+log_level = 0
+Debug = 0
+CustomDebugFile = None
+
+
+def module_loglevel_set(cur_item, cur_value):
```
sthibaul, line 38: this, however, we indeed want to have for `module_process.c` to call it

src/modules/speechd_python_modules/module_utils.py
sthibaul: We can simply call `int()` and return -1 if an exception is caught

src/modules/speechd_python_modules/module_utils.py
```diff
@@ -27,6 +27,58 @@
...
+        CustomDebugFile = new_custom_debug_file
+        if Debug == 1:
+            Debug = 3
+        else:
+            Debug = 2
+    else:
+        if Debug == 3:
+            Debug = 1
+        else:
+            Debug = 0
```
sthibaul, line 73: I don't think we want to mimic module_util's way of emitting debug information. Better use a pythonic way

sthibaul:
> Python output module protocol implementation (module_main and module_process)

As mentioned above, `module_process` can be taken from libspeechd_module. We can just make the python module expose the few C functions `module_*` functions to interact with it. That'll avoid having to maintain several implementations of the core protocol.

daxmawal:
Thanks for the review. I'll be very busy throughout September and October, but I'll get back to this as soon as I can.

PR #1107, opened 2026-08-12
KittenTTS module.

jsett:
# KittenTTS Module.

This is a speech dispatch model for running [Kitten TTS](https://github.com/KittenML/KittenTTS/tree/main). Kitten TTS is a deep learning model which provides high quality natural sounding TTS generation using models ranging from 15M to 80M parameters. Due to its small size it is able to run in real-time on CPU. The goal of this project is to integrate Kitten TTS with speech dispatch while maintaining its near real-time speech generation, with special attention being place on reading of long text's such as ebooks. To achieve this the original python code was rewrote into c and tightly integrated into a speech dispatch model.

There is a lot here so let me give a quick summary of what all is going on here.

## kitten_server.c: 
This file is for handling the protocol and follows some what closely to the example modules for async servers with speech dispatch handling the audio. By design I made sure very little work is done in any function in this thread. Any long running code should be handed to an async queue and ran on one of the threads in the kitten_worker.c file.

## kitten_worker.c:
This is where the handling of long running tasks is done. We create two different threads here one to one to handle passing audio back to the server(since the module_tts_output_server function can block and we want the generation to continue while we are outputting audio to the server). The other thread is dedicated to the generation of audio by the model. We synchronize all this using two GasyncQueue, one for handling incoming speak requests, and the other to handle the outputted audio from our model. we also keep track of how much audio we have generated and played so that long speak commands don't run the cpu unnecessarily hard (For example I have seen that Okular will in some cases send an entire book's text in a single speak command). There is also code here for parsing ssml text using libxml.

## kitten_model.c:
The handles everything we need to do to generated output from onnx using our model. The most important functions here are init_voice_style to handle loading voice styles. reload_models_and_voices: to handle changing the voice style. And kitten_speak for generating audio.

## kitten_downloader.c:
This handles downloading the model+voice styles if its not already on our computer. If it download it does verify the file against a sha256, but that verification is not strongly enforce and is more of a warning.


configure.ac
```diff
@@ -169,6 +169,18 @@ AC_SUBST([SNDFILE_CFLAGS])
...
+AC_SUBST([LIBONNXRUNTIME_LIBS])
+
+PKG_CHECK_MODULES([LIBXML], [libxml-2.0])
+AC_SUBST([LIBXML_CFLAGS])
+AC_SUBST([LIBXML_LIBS])
+
+PKG_CHECK_MODULES([LIBCURL], [libcurl])
+AC_SUBST([LIBCURL_CFLAGS])
+AC_SUBST([LIBCURL_LIBS])
+
```
sthibaul: We want to make these optional, and just build speech-dispatcher without KittenTTS support. See how it is done e.g. for the espeak_ng module, we only build it if espeak-ng is available.

sthibaul: Also, we want to make the CI exercise building the KittenTTS module, i.e. add to `.github/workflows/ci.yml` the apt-get dependencies for these: libonnxruntime-dev, libxml2-dev, libcurl4-gnutls-dev.

src/modules/kitten_server.c
```diff
@@ -0,0 +1,134 @@
...
+
+int module_init(char **msg)
+{
+	fprintf(stderr, "initializing\n");
+
+    init_model_thread_pool();
+
+    *msg = strdup("ok!");
+
+	return 0;
```
sthibaul: We *really* want the init function to report any error that renders the module non-working. Otherwise if the init function returns success, the module which show up as a choice for users, they will try to select it, and then the computer will suddenly be completely silent and the user be at a loss.

src/modules/kitten_model.c
```diff
@@ -0,0 +1,434 @@
...
+    return speed_tensor;
+}
+
+int init_model(const char* model_path){
+
+    // Initialize the ONNX Runtime API table
+    g_ort = OrtGetApiBase()->GetApi(ORT_API_VERSION);
+    if (!g_ort) {
+        fprintf(stderr, "Failed to initialize ONNX Runtime API.\n");
+        return -1;
```
sthibaul, line 292: we do want to report this in the module initialization function

src/modules/kitten_worker.c
```diff
@@ -0,0 +1,567 @@
...
+    xmlBufferFree(buffer);
+    xmlFreeDoc(doc);
+    xmlCleanupParser();
+
+    return output;
+}
+
+/**
+ * Splits a GString into chunks containing at most max_utf8_chars UTF-8 characters.
+ * Returns a GPtrArray of GString pointers containing valid UTF-8 strings.
```
sthibaul, line 304: for much better result, a dividers parameter should be added, to tell where splitting is preferred. This is important to get good prosody: if the splitting is done in the middle of a sentence, the prosody will look very bad. See e.g. module_get_message_part

src/modules/kitten_worker.c
```diff
@@ -0,0 +1,567 @@
...
+}
+
+// this will stop our generation
+// its called from module_pause and module_stop
+int model_stop_generation(){
+    stop_set(true);
+    return 0;
+}
+
+// This is the same as spd_pthread_create, I move it in here because including $(common_SOURCES) was creating circular dependences for me.
```
sthibaul: What kind of circular dependences were showing up? We should be able to fix them rather than duplicating code.

src/modules/kitten.h
```diff
@@ -0,0 +1,133 @@
+#include <glib/gstdio.h>
```
sthibaul, line 13: We need a copyright header at the beginning of every added file, to specify the licencing of the file and the copyright holders

src/modules/kitten.h
```diff
@@ -0,0 +1,133 @@
...
+        } \
+    } while (0)
+
+extern int model_type;
+extern int ROWS;
+#define COLS 256
+
+#define PAD "$"
+#define PUNCTUATION ";:,.!?¡¿—…\"«»\"\" "
+#define LETTERS "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
```
sthibaul, line 54: does this TTS only support english? Can it support more in the future?

src/modules/kitten_downloader.c
```diff
@@ -0,0 +1,196 @@
...
+    g_mapped_file_unref(mfile);
+
+    return match ? 0 : -1;
+}
+
+// downloads the models and voices if they do not already exist.
+// also verifys using sha256 and checks file size.
+int download_models(void) {
+    // Build absolute destination directory path: ~/.config/speech-dispatcher/extra/
+    char *target_dir = g_build_filename(home_dir, TARGET_SUBDIR, NULL);
```
sthibaul: it should rather go to `.cache/speech-dispatcher`, and I'd say KittenTTS rather than an non-informative "extra"

src/modules/kitten_server.c
```diff
@@ -0,0 +1,134 @@
...
+    return 1; // delivery to the synthesizer is successful
+}
+
+size_t module_pause(void)
+{
+	/* Pause playing */
+	fprintf(stderr, "pausing\n");
+
+    // does not seem there is a resume function so
+    // pause will be handled the same as stop.
```
sthibaul, line 212: as mentioned in the module output documentation, pause is indeed similar to stop, but the difference is that it is preferrable to stop at a place where the indexing position is available, so that if the screen reader brings the cursor to that position, the feedback to the user is coherent with that positionning: not before, not after, exactly where the audio feedback finished.

sthibaul: (of course, that's a refinement that we can integrate later, doing the same as stop is fine for now)

sthibaul (changes requested):
Thanks for the nice work!

In the future, we will probably want to integrate various onnx-based voices, so code will be useful to share between modules, but your writing seems already quite well structured so that it will be convenient to do.

There are just a few changes that need to happen before we can integrate this.

configure.ac
jsett: I should be able to fix this.

src/modules/kitten_server.c
jsett: This should not be to hard to fix.

src/modules/kitten_model.c
jsett: same thing.

src/modules/kitten_worker.c
jsett: I am going to push back on this one. This is chunking function is to handle somewhat of a edge case. Basically most of the chunking it done when parsing the SSML. Which gives us nice bite sized sentences. And we actually want to speak the entire sentence when we can and not break it into splits as this produces the best output. The edge happens when a sentence is over 399 char's and the model can't handle that long of a text. For the most part 399 char's in a single sentence is pretty rare, because this is almost the same char count as 3-5 average sentences. So in short most of the time we will not be doing any chunking and giving the user options to configure where they want splitting would likely just create confusion not to mention add complexity for a edge case that rarely happens.

src/modules/kitten_worker.c
jsett: So, I can see that cicero and dummy both use spd_pthread_create successfully and they build fine but when I try and model my compile flags and includes to match I get errors. Either stuff has "undefined reference" or "multiple definition" depending on what I try. It could be some issue with how I structured my .h file or something is sneaking references into the build files. To be honest I really don't know, It's very confusing for me. This seems to be the only function that if I try and use it cause issues. I will try looking at it again, but I really don’t know what causing this issue.

src/modules/kitten_worker.c
sthibaul: > The edge happens when a sentence is over 399 char's

Ah, ok, then that's fine!

src/modules/kitten.h
jsett: I think this model only supports English. But I think that the architecture it uses supports multiple languages. According to the [StyleTTS2](https://github.com/yl4579/StyleTTS2) readme "Non-English dataset: You can train on any language you want". Also [Kokoro](https://huggingface.co/hexgrad/Kokoro-82M) which is very similar and also based off StyleTTS2 doe support multiple languages according to their [docs](https://huggingface.co/hexgrad/Kokoro-82M/blob/main/VOICES.md) (Of course Kokoro is a much larger almost 4x as large ). But it would not surprise me if with minimal changes Kokoro could be supported. I choose this model primarily because of its size and my primary goal was to have real-time generation on CPU.

src/modules/kitten_downloader.c
jsett: This should be an easy change.

src/modules/kitten.h
jsett: For the code I wrote which are all the kitten_* files, I will probably place them under a MIT License. My primary desire is to keep the stuff that may be reusable on another project in a highly  permissive license for example kitten_model.c very well may be reusable else where. I don't think this would cause any with speechd, even thought its under GPL, but let me know if that is an issue. Also note I am not the copyright owner for the model. According to its [hugging face page](https://huggingface.co/KittenML/kitten-tts-mini-0.8) Its under the apache-2.0.

sthibaul: Indeed, MIT will be completely fine.

> Also note I am not the copyright owner for the model

Sure, that's not a problem, people will be able to download them fine.

BTW, I'm however thinking: distributions will probably like to be able to install the models from packages, so they are already there and users don't have to download them before use. It would thus be useful to also try to look for the models somewhere in `/usr/share/speech-dispatcher`.

configure.ac
jsett: I have updated the code to fix this.

src/modules/kitten_server.c
jsett: I have updated the code to fix this.

src/modules/kitten_model.c
jsett: I have updated the code to fix this.

src/modules/kitten_worker.c
jsett: So I had some time to look at this and I don't think this is an issue with my code. It seems that its probably something with module_utils.h. The other modules that include spd_pthread_create such as dummy.c, get away with it because they don't have the module_loop and module_* functions that seem to get multiple definitions.
For example you can break dummy.c and recreated this issue by adding in
```
int module_debug(int enable, const char *file)
{
	return 0;
}
```

Fixing this would likely require changing module_utils which is outside the scope of what am working on. So I don't have a solution other then coping in spd_pthread_create.

src/modules/kitten.h
jsett: I have updated the headers with the license.

jsett: I also added code to check if the models are in `/usr/share/speech-dispatcher/models/kitten` and use that path if they are.

src/modules/kitten_downloader.c
jsett: I updated this.

src/modules/kitten_downloader.c
unlimitedsola, line 1: I believe this is best to be handled by package managers instead?

As models update, we don't want to frequently update and recompile this as downstream packages might lag behind for years.

unlimitedsola: Also all the download links are outside of this project's control. Which means that we won't be able to fix a broken download until a new release is packaged by distributions and shipped to the end user (which as commented, might be months/years after).

src/modules/kitten_server.c
jsett: If by indexing position you mean the module_report_index_mark, Then this is already done.

src/modules/kitten_downloader.c
jsett: > I believe this is best to be handled by package managers

It already checks if the models/voices are in the path `/usr/share/speech-dispatcher/models/kitten` and uses that path first. If the package manger placed them there, then they will get used.

> we won't be able to fix a broken download until a new release

I have added the ability to set the download configuration through the models dot conf file. 

Here is an example dot conf on how to do that https://gist.github.com/jsett/6146eb2803f8780a1830ed816bdc94ef

unlimitedsola: > > I believe this is best to be handled by package managers
> 
> It already checks if the models/voices are in the path `/usr/share/speech-dispatcher/models/kitten` and uses that path first. If the package manger placed them there, then they will get used.

I'm afraid distribution would not organize their packages this way.

Like piper, kitten models are general purpose in the sense of the Text-to-Speech task. If we take example from piper, the AUR packages place the models into `/usr/share/piper-voices/`, because their use are not limited to speech-dispatcher, and the distributions would not want to package multiple copies of large model files just for different applications to use.

It is also likely that different distributions wouldn't come to agreement on where these models should be placed. So the distribution packagers would adjust the conf default to accommodate their packaging schemes.

> I believe this is best to be handled by package managers instead?

Nevertheless, I personally wouldn't expect program like speech-dispatcher (as a daemon process) to make network requests (and I strongly believe that we should learn from the log4shell disaster where a logging library contains the code path to make network requests) . So adding this to me feels like a scope creep that I would like to push against, as indicated by the first time linking with libcurl.

src/modules/kitten_worker.c
jsett: I created a bug report for this since it happens with multiple different modules. https://github.com/brailcom/speechd/issues/1111

src/modules/kitten_downloader.c
jsett: `/usr/share/speech-dispatcher` was recommended in the comments above, but I really don't have any real issue changing it if that what is decided.

> code path to make network requests

I'm not opposed to removing the downloader. But there is a trade off here. If you remove it you increase the complexity for users to get things working, if there distro did not include the models with the install. I would be interested in hearing what other people think.

unlimitedsola: > If you remove it you increase the complexity for users to get things working, if there distro did not include the models with the install.

We can document what distro package the user should install to get the models in the default conf template (the distro maintainers can add the documentation themselves too). Or we could also link to one-off scripts to help users to download the models into their home directories.

Personally, what matters to me is that the downloader should be a separate program/script from the daemon process code, so it is only loaded when the user explicitly request to perform the one-off download operation.

jsett: Turning the downloader into a one-off script might not be a bad idea. I would like to hear from @sthibaul since i don't know what the most convenient way of doing this for speachd would be. Like would I just need to add a extra build target or something else. Also I could pretty easy rewrite the downloader as a python script or something but again, I don't know if that would be better or worse for packaging for the distros. Frankly I'm not super experienced on the intricacies of packaging.

sthibaul: > If we take example from piper, the AUR packages place the models into /usr/share/piper-voices/, because their use are not limited to speech-dispatcher, and the distributions would not want to package multiple copies of large model files just for different applications to use.

That is right.

> It is also likely that different distributions wouldn't come to agreement on where these models should be placed.

Well, if upstream of these models could simply document where these are supposed to be shipped, then "magically" all distributions would come to agreement...

If upstream does not do that, it's a "matter" of settling a de-facto standard. Is there some distribution or software that already puts these somewhere which is software-agnostic?

src/modules/kitten_server.c
sthibaul: yes, it's the index mark reporting

sthibaul:
bleh, "ubuntu-latest" doesn't seem to really be "latest", the CI is taking 24.04 rather than 26.04... Let me see that

sthibaul:
> bleh, "ubuntu-latest" doesn't seem to really be "latest", the CI is taking 24.04 rather than 26.04... Let me see that

Could you rebase? I have bumped to 26.04 so onnxruntime is available.

src/modules/kitten_worker.c
sthibaul: I have fixed it, you should now be able to just use spd_pthread_create without having to take module_utils

src/modules/kitten_downloader.c
sthibaul: > I personally wouldn't expect program like speech-dispatcher (as a daemon process) to make network requests

I tend to agree with that, also because the first speech request will take time, making the user wonder what is happening.

sthibaul: > Turning the downloader into a one-off script might not be a bad idea

That could be an idea yes. People installing speech-dispatcher by hand would use it, and distribution people would package the voices. People could still use the script to update the voices.

Again, if only upstream voices providers could manage these questions instead of just throwing .onnx files at people...

There could simply be a voice downloading manager, even independent from speech-dispatcher, that manages updates and whatnot, and speech-dispatcher would just happen to see the voice files show up, and use them.

jsett: > upstream of these models could simply document where these are supposed to be shipped

Unfortunately I don't think hugging face models have a way of doing this currently.

jsett: Based off the above comments, I think this is what I am going to do.

With regards to the downloader.

1. Remove the downloader from the speechd module.

2. Create a separate script outside speechd code that has the downloader (this will likely be in python).

3. Add documentation on how to use the downloader.


With regards to the search path for the distro's.

1. Search if the models are in `~/.cache/speech-dispatcher/kitten` (for user provided models).

2. Check if a path is provided in the module's .conf file. Something like a line for `modelsPath "<path>"` distro's then should be able to place the models where ever they desire and add a line in the conf file.

sthibaul: > > upstream of these models could simply document where these are supposed to be shipped
> 
> Unfortunately I don't think hugging face models have a way of doing this currently.

It would really just be a matter of documenting it in README or website or whatever, nothing fancy...

sthibaul:
You need to add `kitten.h` to `EXTRA_DIST` in `Makefile.am` so it gets shipped in the distributed tarball.

src/modules/kitten_downloader.c
jsett: I made all those changes in the latest commit.

configure.ac
```diff
@@ -425,6 +426,29 @@ AM_CONDITIONAL([kali_support], [test $with_kali != no])
...
+AS_IF([test $with_kitten != "no"],
+  [PKG_CHECK_MODULES([LIBONNXRUNTIME], [libonnxruntime],
+      [kitten_libonnxruntime="yes"],
+      [AS_IF([test $with_kitten = "yes"],
+        [AC_MSG_FAILURE([libonnxruntime is not available])])])
+   PKG_CHECK_MODULES([LIBXML], [libxml-2.0],
+      [kitten_libxml="yes"],
+      [AS_IF([test $with_kitten = "yes"],
+        [AC_MSG_FAILURE([libxml-2.0 is not available])])])
+   AS_IF([test "x$kitten_libonnxruntime" = "xyes" -a "x$kitten_libxml" = "xyes" -a "x$kitten_libcurl" = "xyes"],
```
sthibaul: You need to drop the libcurl condition, and you can drop it from the github workflow

src/modules/README.kitten.md
```diff
@@ -0,0 +1,23 @@
+# Module conf file options.
```
sthibaul, line 1: Please rather put an example configuration file, otherwise it will be too hard for users to work out what it should look like. You can then stuff this information in the example configuration file.

src/modules/README.kitten.md
```diff
@@ -0,0 +1,23 @@
...
+The modules dot conf file can be used to configure the module using `AddVoiceFile` and `modelPath`. `AddVoiceFile` let you set models/voices in the case of updated models. `modelPath` let you set the modes/voices search path.
+
+## AddVoiceFile
+
+AddVoiceFile should be formated like such.
+```
+AddVoiceFile "voice_quality" "model filename" "model sha256" "voices filename" "voices sha256"
+```
+
+you can find an example config [here](https://gist.github.com/jsett/b82d19d3e2e538a29e5e79995b619a62)
```
sthibaul: Better inline the config in the example config file :)

configure.ac
jsett: I must have over looked that. Interesting that it did not cause my ./configure to fail.

src/modules/README.kitten.md
```diff
@@ -0,0 +1,23 @@
+# Module conf file options.
+
+The modules dot conf file can be used to configure the module using `AddVoiceFile` and `modelPath`. `AddVoiceFile` let you set models/voices in the case of updated models. `modelPath` let you set the modes/voices search path.
```
sthibaul, line 3: You can set the model in the example file to /usr/share/kitten, to give people an idea where the files should be put as a distribution (to be shared with other software using kitten models)

sthibaul (changes requested):
we're getting close :)

configure.ac
sthibaul: ./configure is meant to avoid failing unless the user explicitly requested --enable-kitten

src/modules/README.kitten.md
jsett: I updated the readme with examples.

src/modules/README.kitten.md
```diff
@@ -18,6 +18,18 @@ modelPath should be formated like such.
...
+## Example configuration file
+
+`kittentts.conf`
+
+```
+modelPath "/usr/share/kitten"
+
+AddVoiceFile "Normal" "kitten_tts_micro_v0_8.onnx" "95481626fee1ba70ce683e69c534fc7cb38433c46ce42d3abbeafb4b9f1a4123" "voices_micro.bin" "12ad10f1fcce8a458b5cf79769b8edd4ba0e11e9fb6532fd192c3500b2b37a5d"
+AddVoiceFile "High" "kitten_tts_mini_v0_8.onnx" "0f5bbae4fc4800c98dbc544a87ecfa79510de2fb8222db30d12e5bfe9177df91" "voices_mini.bin" "0e4965b46333db53ce09c73842623bf7055ea62c67f78803cb7ea9c16da6ac2b"
+AddVoiceFile "Low" "kitten_tts_nano_v0_2.onnx" "42fa8809db319cd7c4c83b3c501e2313bf90edf610235291cad605e4adcb242d" "voices_nano.bin" "42a40a24a352a38657d6cb86ceee51bbc2b7780b29e04fb60bcdf959adccea01"
```
sthibaul, line 44: Better put it as an actual file in `config/modules` (and make it installed from Makefile.am), so it's really completely ready for distributions to just use it without having to dive into document etc. and people will actually see this documentation when reading the configuration files by inlining it in it.

.github/workflows/ci.yml
```diff
@@ -144,6 +145,7 @@ jobs:
                         libdotconf-dev libglib2.0-dev libltdl-dev- libsystemd-dev systemd
                         gettext help2man texinfo texlive texlive-plain-generic
                         pulseaudio
+                        libonnxruntime-dev libxml2-dev libcurl4-gnutls-dev
```
sthibaul: Please remove the libcurl dependency,  now unused :)

src/modules/kitten_server.c
```diff
@@ -0,0 +1,226 @@
...
+#include "kitten.h"
+
+#include <dotconf.h>
+
+static DOTCONF_CB(cb_voicefiles);
+static DOTCONF_CB(cb_modelPath);
+
+static const configoption_t options[] = {
+	{"AddVoiceFile", ARG_LIST, cb_voicefiles, NULL, CTX_ALL},
+	{"modelPath", ARG_STR, cb_modelPath, NULL, CTX_ALL},
```
sthibaul, line 25: Better add a `Debug` option, like the other modules, and make use of `MSG` and `DBG` instead of always printing to stderr, so people don't have a log file growly a lot, but can still easily add debugging output when something goes wrong and you need information from the execution on the user machine to investigate.

src/modules/README.kitten.md
jsett: added config/modules/kittentts.conf

.github/workflows/ci.yml
jsett: Should be removed now.

src/modules/README.kitten.md
sthibaul: Really, rather put the information of README.kitten.md in the configuration file example. That's the best way that users will see it and know how to configure it. Otherwise distributions will at worse not install the readme file, at best put it in /usr/share/doc/speech-dispatcher, where users won't find it, and be at a loss when trying to configure /etc/speech-dispatcher/config/kitten.conf

sthibaul:
The CI said:

```
kitten_worker.c:416:57: warning: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘size_t’ {aka ‘long unsigned int’} [-Wformat=]
```

This indeed should rather be a `%zd`, to properly print a `size_t`

src/modules/kitten_server.c
jsett: I'm confused on what we are talking about here. Are we talking about replacing the `fprintf(stderr,` functions(there in quite a few locations) with MSG/DBG. Or is this something that is added to the dot conf stuff?

sthibaul:
The CI also said: 

```
[8](https://github.com/brailcom/speechd/actions/runs/33558231873/job/100024779175?pr=1107#step:9:149)
kitten_worker.c: In function ‘parse_ssml_to_gqueue’:
kitten_worker.c:274:13: warning: ‘content’ is deprecated [-Wdeprecated-declarations]
  274 |             pl->text = g_string_new(buffer->content);
      |             ^~
In file included from /usr/include/libxml2/libxml/parser.h:20,
                 from kitten.h:28,
                 from kitten_worker.c:13:
/usr/include/libxml2/libxml/tree.h:115:14: note: declared here
  115 |     xmlChar *content XML_DEPRECATED_MEMBER;
      |              ^~~~~~~
kitten_worker.c:285:5: warning: ‘use’ is deprecated [-Wdeprecated-declarations]
  285 |     if (buffer->use > 0) {
      |     ^~
/usr/include/libxml2/libxml/tree.h:121:18: note: declared here
  121 |     unsigned int use XML_DEPRECATED_MEMBER;
      |                  ^~~
kitten_worker.c:289:9: warning: ‘content’ is deprecated [-Wdeprecated-declarations]
  289 |         pl->text = g_string_new(buffer->content);
      |         ^~
/usr/include/libxml2/libxml/tree.h:115:14: note: declared here
  115 |     xmlChar *content XML_DEPRECATED_MEMBER;
      |              ^~~~~~~
```

The xml header mentions that one should rather use `xmlBufferContent` and `xmlBufferLength`.

src/modules/kitten_server.c
sthibaul: yes, *and* add a Debug dot conf option that sets the Debug variable, that the MSG/DBG calls will use to determine whether they should actually print their content or not. That is *very* convenient to get users to provide debugging information easily

jsett: I made these changes.

jsett:
Both of those should be fixed now.

src/modules/README.kitten.md
sthibaul: I seems I don't manage to get my point understood :)

People won't find this readme file. But the people who do edit configuration files will  find the kittentts.conf file (you just need to add the file name in config/Makefile.am to get it installed on users' system), so all the content of this readme file should really be in kittentts.conf for people to find it

src/modules/Makefile.am
```diff
@@ -133,6 +134,16 @@ sd_flite_LDADD = $(top_builddir)/src/common/libcommon.la \
 	$(common_LDADD)
 endif
 
+#
+# kitten tts module
+#
+if kitten_support
+modulebin_PROGRAMS += sd_kittentts
```
sthibaul, line 141: do we really want to call it "kittentts" rather than just "kitten"? The "tts" part seems redundant to me, we already know that it is a tts

config/modules/kittentts.conf
```diff
@@ -0,0 +1,5 @@
+modelPath "/usr/share/kitten"
+
+AddVoiceFile "Normal" "kitten_tts_micro_v0_8.onnx" "95481626fee1ba70ce683e69c534fc7cb38433c46ce42d3abbeafb4b9f1a4123" "voices_micro.bin" "12ad10f1fcce8a458b5cf79769b8edd4ba0e11e9fb6532fd192c3500b2b37a5d"
+AddVoiceFile "High" "kitten_tts_mini_v0_8.onnx" "0f5bbae4fc4800c98dbc544a87ecfa79510de2fb8222db30d12e5bfe9177df91" "voices_mini.bin" "0e4965b46333db53ce09c73842623bf7055ea62c67f78803cb7ea9c16da6ac2b"
+AddVoiceFile "Low" "kitten_tts_nano_v0_2.onnx" "42fa8809db319cd7c4c83b3c501e2313bf90edf610235291cad605e4adcb242d" "voices_nano.bin" "42a40a24a352a38657d6cb86ceee51bbc2b7780b29e04fb60bcdf959adccea01"
```
sthibaul, line 5: As mentioned in the comments for readme, please put the content of the readme in here as comments, so people directly have everything at hand when configuring. See the other configuration files.

And that includes the debug line, to be available here ready for getting enabled by users who happen to have issues.

sthibaul (changes requested):

src/modules/Makefile.am
jsett: I updated this.

config/modules/kittentts.conf
jsett: I added comments into the conf to document the conf options and added it to the makefile.am

jsett:
The conf file has been updated with comments on the configure options.

src/modules/README.kitten.md
sthibaul: You can now remove this readme file, since its content is fully moved to the configuration file

jsett: removed the readme.

sthibaul:
Thanks!

Issue #1092, opened 2026-05-03
cxxpiper improvement


sthibaul:
https://github.com/brailcom/speechd/pull/996 introduced a cxxpiper module, but it needs improvement, as mentioned in the comments, but which have been unanswered, so copy/pasting them here so they are not lost:

- it will be nice long-term to be able to load several models. Even better, list all of them but load them only on-demand, so people don't have to modify any file but just install a model package and see the language pop-up in orca
- It's quite sad that piper doesn't provide parsePhonemizeConfig, parseSynthesisConfig, parseModelConfig, initialize, loadModel, loadVoice, synthesize, textToAudio,  as library functions. It'd be good to tell upstream to do so, rather than copying their code
- we'd want to see piper support pipelining, to start sending audio to the server before the whole audio is produced. That would be really important to get good reactivity
- It would be useful to make CUDA support work

sthibaul:
> we'd want to see piper support pipelining

Unfortunately, looking at the onnx model and thinking about it, it's not possible, convolutions etc. make partial evaluation essentially not possible. The existing sentence-by-sentence split of speechd is the best we'd have without affecting the prosody.

sthibaul:
> It would be useful to make CUDA support work

I tried it, it depends on having the rttensor etc. stack installed, not very easy, but should be working fine. That should help a *lot* with the synthesis latency.

sthibaul:
> It's quite sad that piper doesn't provide parsePhonemizeConfig

Apparently piper_phonemize https://github.com/rhasspy/piper-phonemize/ was reworked last year and https://github.com/OHF-Voice/piper1-gpl would have to be used nowadays. I don't know if they migrated the pull requests and issues...

PR #1108, opened 2026-08-15
Rework cxxpiper module

unlimitedsola:
Disclaimer: Code changes assisted by LLM. I have very limited experience with C++ and autotools. I've done my best to review the code and verify that it at least works on my machine. Apologies for any obvious mistakes that I didn't catch.

---

Migrate now deprecated/archived piper-phonemize[1] based cxxpiper implementation to libpiper[2] which was recently added to the upstream piper repository. (as requested in https://github.com/brailcom/speechd/issues/1092#issuecomment-4382164999)

Because we are now using the upstream libpiper C library directly, I figured we should probably also remove the cxx prefix from the module name. (on top that, also many breaking changes and redesigns)

Changes made:

- Removed conf options that new libpiper no longer support
- Added peak normalization for parity with Python CLI
- Simplified voice handling and configuration

[1]: https://github.com/rhasspy/piper-phonemize
[2]: https://github.com/OHF-Voice/piper1-gpl/tree/main/libpiper


sthibaul:
I added CI on top of this to be able to check the build, and the code indeed looks fine and much more maintainable :)

sthibaul:
Thanks!

PR #1110, opened 2026-08-25
Add Edge TTS generic output module

Jyatin:
## Summary

Adds an initial generic output module configuration for using Microsoft Edge TTS with Speech Dispatcher.

### Changes

- Add `config/modules/edgetts-generic.conf`
  - Adds support for `en-US`, `hi-IN`, `bn-IN`, and `gu-IN`
  - Adds several Edge TTS neural voices
  - Configures the generic module to invoke an Edge TTS wrapper
- Change the generic module shell from `/bin/sh` to `/bin/bash`
  - Required because the generic module currently executes commands using `set -o pipefail`
  - `/bin/sh` resolves to `dash` on the tested WSL environment, which does not support `pipefail`

### Testing

Tested locally on WSL2 with Speech Dispatcher and Edge TTS.

Verified that:

- Speech Dispatcher invokes the generic module.
- The Edge TTS wrapper is invoked successfully.
- Edge TTS generates an MP3 file.
- Audio playback works successfully.


config/modules/edgetts-generic.conf
```diff
@@ -0,0 +1,26 @@
+Debug 1
```
sthibaul: Please add copyright information, like other .conf files

config/modules/edgetts-generic.conf
```diff
@@ -0,0 +1,26 @@
...
+
+AddVoice "bn-IN" "MALE1" "bn-IN-PrabhatNeural"
+AddVoice "bn-IN" "FEMALE1" "bn-IN-TanishaaNeural"
+
+AddVoice "gu-IN" "MALE1" "gu-IN-NiranjanNeural"
+AddVoice "gu-IN" "FEMALE1" "gu-IN-DhwaniNeural"
+
+DefaultVoice "en-US-EmmaNeural"
+
+GenericExecuteSynth "/root/bin/speechd-edge-wrapper \"$DATA\" >/dev/null 2>&1"
```
sthibaul: is speechd-edge-wrapper really supposed to be in /root/bin??

Also, please use GenericCmdDependency to make sure to automatically disable the module when the command is not available.

Are all voices always available, or does that depend on the existence of a file? If the latter, please use VoiceFileDependency so that voices are shown to the user only when available.

sthibaul (changes requested):
Thanks for this, just a few fixes are needed.

config/modules/edgetts-generic.conf
unlimitedsola: Also what provides this `speechd-edge-wrapper`? I searched online and couldn't find anything.

Jyatin:
Thanks for the review. I’ve addressed the requested changes:

Added copyright information to the configuration.
Replaced the unavailable speechd-edge-wrapper with the edge-tts command directly.
Added GenericCmdDependency "edge-tts" so the module is disabled when the command is unavailable.
Verified the configuration locally and confirmed that Edge TTS successfully generates audio.

Please let me know if any further changes are needed.

Jyatin:
he Edge TTS voices are provided by the online Edge TTS service and do not depend on local voice files, so VoiceFileDependency is not applicable here. GenericCmdDependency "edge-tts" ensures that the module is disabled when the edge-tts command is unavailable.

config/modules/edgetts-generic.conf
```diff
@@ -0,0 +1,43 @@
...
+# and the generic module to play the resulting audio.
+
+# Debug turns debugging on or off
+# See speechd.conf for information where debugging information is stored
+Debug 0
+
+GenericDefaultCharset "utf-8"
+
+GenericExecuteSynth \
+"printf %s \'$DATA\' >$TMPDIR/edge-tts.txt && edge-tts -f $TMPDIR/edge-tts.txt -v \'$VOICE\' --write-media $TMPDIR/edge-tts.mp3 && $PLAY_COMMAND $TMPDIR/edge-tts.mp3"
```
sthibaul: Unfortunately, when the user configured to use the alsa audio device, $PLAY_COMMAND is `aplay`, which does not support the mp3 formation. Can edge-tts support wave format? That'll also avoid having to compress the audio, only for the audio player to re-decompress it.

Also, ideally you would avoid using temporary files, by passing the data through pipes `|` . If `edge-tts` does not have a special notation for that (it's often `-` to be given as file name), you can trick it by passing it `/dev/stdin` and `/dev/stdout`.

sthibaul (changes requested):
You also need to add the file in `config/modules/Makefile.am` to `dist_moduleconf_DATA` and `dist_moduleconforig_DATA`

sthibaul:
Also, please mention the support in `README.md` ;)

Jyatin:
Thanks for the feedback. I’ve updated the Edge TTS module to avoid temporary files and use a pipe-based pipeline. The MP3 output from edge-tts is now converted to WAV using mpg123 before being passed to $PLAY_COMMAND.

I’ve also added mpg123 as a GenericCmdDependency, added edgetts-generic.conf to both distribution lists in Makefile.am, and documented Edge TTS support in README.md.

git diff --check passes and the working tree is clean. I wasn’t able to test the complete audio pipeline locally because mpg123 and aplay aren’t installed in my current WSL environment.

sthibaul:
Thanks for this!

it'll still be useful to add support for rate and pitch :) 
(you'll most probably need  to use the Generic{Rate,Pitch}{Multiply,Add} options to  perform a conversion)

PR #1117, opened 2026-09-01
Add fono generic output module

bogdanr:
Adds a generic output module config for [fono](https://fono.page), a command-line voice stack.

`fono speak stream` reads text on stdin and plays the audio itself, so the module is a one-liner and needs no `$PLAY_COMMAND`. It stops on SIGKILL, so stop and cancel work.

Which synthesizer actually speaks is fono's business, not this file's: on-device ONNX (Kokoro, Piper, Supertonic), a Wyoming server, an OpenAI-compatible endpoint, or a cloud vendor like ElevenLabs or Cartesia. Switching backends does not touch the config here.

Voices are mapped by fono's palette labels (`male 1`, `female 2`, ...) instead of raw model ids, so the `AddVoice` lines hold across backends. `DefaultVoice "auto"` lets fono pick per the language of the text.

Same shape as `edgetts-generic.conf` and `mimic3-generic.conf`: config only, no build changes, `GenericCmdDependency "fono"` so the module is skipped when fono is not installed.

Tested on Slackware with speech-dispatcher 0.12.1: `spd-say -o fono-generic`, voice selection with `-y`, and `spd-say -C` mid-utterance.


sthibaul:
Thanks!

